PR #8, opened 2017-12-07
Add support for sending JSON data and path element data.

cpubob:
This is mainly to support Nokia who's beta software is still using GNMI 0.3.1.  Adding option to allow sending path data in "element" format and check for user provided JSON data directly on CLI.

### DIFF
--- a/gnmi_set/gnmi_set.go
+++ b/gnmi_set/gnmi_set.go
@@ -47,15 +47,17 @@ func (i *arrayFlags) Set(value string) error {
 }
 
 var (
-	deleteOpt  arrayFlags
-	replaceOpt arrayFlags
-	updateOpt  arrayFlags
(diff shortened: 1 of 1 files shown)


samribeiro:
I am a bit reluctant to add deprecated functionalities to this. This is meant to be used as a reference implementation where somebody can go and read the code. Having to read through code for deprecated functionalities is consuming. However, I am assuming this is going to save somebody's time so I see no reason to block this. But I would request a few changes:
- create a deprecated module under utils that gnmi_set will import. Move the deprecated flag there. Prepend it with "deprecated". Move all the json helpers there. In sum minimize the code changes to gnmi_set.go as much as possible. Thanks!

cpubob:
I agree that adding deprecated options is suboptimal.  I am totally fine just keeping these changes in my personal repo since its only to help me and eventually Nokia will be on gnmi 0.4.0.  I am ggoing to cancel this pull request.

Thanks for taking the time to review.

PR #13, opened 2018-03-08
Fix readmes and openflow.json

mike-albano:
(M) gnmi_target/README.md
-fix the run example (add missing ''\")

(M) README.md
-update main README to include go version requirement.

(M) gnmi_target/openconfig-openflow.json
-Update JSON to include a "state" container. (gnmi_get example fails without this). If this is not appropriate, alternative would be change the gnmi_get
 example to use a path of: "/system/openflow/controllers/controller[name=main]/connections/connection[aux-id=0]" (eg leave off the '/state/address')

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,7 +33,7 @@ These instructions will get you a copy of the project up and running on your loc
 
 ### Prerequisites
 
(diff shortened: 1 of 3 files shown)


samribeiro:
thanks!

PR #37, opened 2018-11-19
Added support for Origin for both GET and SET.

phani2410:
Tested. Origin is parsed from xpath string if present.


phani2410:
Please check. I have accepted the CLA

phani2410:
wrong commit mail

PR #38, opened 2018-11-19
Added support for parsing the origin in the Path.

phani2410:
Checked for ':' in path attribute


phani2410:
Please check travis CI logs. I am unable to figure out. if the issue is due to the code.

samribeiro:
@boleifu can you check if this works for you?

samribeiro:
thanks @phani2410 
@boleifu has updated the go version for the tests, that was likely the cause they were breaking.
I have requested @boleifu to check that this change does not impact his projects.

samribeiro:
may be causing #43

PR #41, opened 2018-12-11
Add Python gNMI client

mike-albano:
Current supported gNMI features:
- GetRequest
- Target hostname override
- Auto-loads Target cert from Target if not specified
- User/password based authentication
- Certificate based authentication

Current unsupported gNMI features:
- Subscribe
- SetRequest
- Multiple xpaths


inetAnt:
Cool stuff !

Issue #43, opened 2018-12-20
x-path with ipv6 address is broken


pradeepvdevan:
route[prefix=2500::2/128] is sent as route[prefix=2500" as ":" causes it to skip or remove rest of it

samribeiro:
Is this caused by #38 ?

pradeepvdevan:
@samribeiro Yes it is caused by support of origin

PR #49, opened 2019-01-29
Added suport for origin

phani2410:
Added support for parsing origin from xpath. 

Purpose is same as https://github.com/google/gnxi/pull/38.

But this change is made to avoid https://github.com/google/gnxi/issues/43.


phani2410:
Wrong email

PR #54, opened 2019-02-08
	modified:   py_gnmicli.py

olivier-leduc:
Adding the notls option for the python gnmi client (might have some use in isolated lab environments).


olivier-leduc:
I signed it!

PR #57, opened 2019-02-08
Fix certificates handling for various Target cert situations.

mike-albano:
Better deal with self-signed Target certificates (including option to get it automatically) and Target certificates signed by public CA.
Fix README examples.


mike-albano:
PRs showing hostname not git.user/email.

PR #60, opened 2019-02-16
bug fix on the notls option

olivier-leduc:


mike-albano:
Need a better description next time. Thanks for fixing.

PR #61, opened 2019-02-18
 send provided client certs always to server

idefixcert:
In my opinion insecure means don't validate server name in certificate.
But we have to provide the certificates to the server.


samribeiro:
Hi @idefixcert,
I agree with you. However, the concern here is that we dont have the proper definition of "insecure". gNMI/gNOI certainly does not have this definition. In fact this implementation supports some attempt at an insecure channel because some Target implementations out there do not use certificates and some users prefer not to complicate when running or testing in lab environments. Because of this and to avoid breaking functionality for the existing implementations, I am going to choose leaving this definition of insecure in place. I hope you understand this. Thanks!

PR #78, opened 2019-07-29
Removing pathlen information in accordance with RFC5280 4.2.1.9

alisharif2:
According to https://tools.ietf.org/html/rfc5280 the pathlen information must be omitted if the CA boolean is false. In this case, since you never assert the CA to be true (hence default false) you must remove the pathlen information.


alisharif2:
I signed it!

samribeiro:
Thanks Ali,
Can you also remove the global variable at the top (line 37):
certMaxPathLen = 5

alisharif2:
I've had to make additional changes after I removed certMaxPathLen variable because I noticed that pathlen information was being placed in other certs. So to summarise:
Template() - Since these aren't CA certs they can't have this info
TemplateCA() - Add pathlen info here
FromSigningRequest() - Since this function is part of the signer() func passed to install and rotate RPCs used to create certs from device-generated CSRs, I've removed the pathlen information since it isn't installing certs with CA:true

samribeiro:
Thanks!

PR #79, opened 2019-09-24
Backup perf experiment scripts for gateway machine 

river8:
This is a temporary solution to back up perf experiment scripts for gateway machines. Gateway machines are intended to be non google machine. So, it is better to back up these scripts to github.


river8:
@googlebot I signed it!

river8:
@googlebot I signed it!

river8:
@googlebot I signed it!

PR #90, opened 2020-04-30
Regenerate generated.go from yang files.

mpsalisbury:
The existing tests in gnmi/ fail. Regenerating from the yang files fixes the problem.


samribeiro:
Thanks!

Issue #97, opened 2020-06-30
Create gNOI Reset Client


loriscn12:
This issue consists of building client for testing the gNOI Reset service implementation.

As part of this client implementation, the following gNOI Reset service operations will be tested:

Start a Factory Reset

PR #102, opened 2020-07-01
gNOI Client

Jared-Mullin:
Issue #97 
Initial client setup in gNOI library, test for ResetError responses included


gnoi/reset/client.go
```diff
@@ -0,0 +1,71 @@
...
+	}
+	return CheckResponse(out)
+}
+
+// CheckResponse checks for errors.
+func CheckResponse(res *pb.StartResponse) error {
+	switch res.Response.(type) {
+	case *pb.StartResponse_ResetSuccess:
+		return nil
+	case *pb.StartResponse_ResetError:
```
samribeiro, line 55: I believe it is better to assume that all errors can be triggered at the same time, and therefore report on them all instead of just the first one caught.

gnoi/reset/client_test.go
```diff
@@ -0,0 +1,70 @@
...
+func TestStart(t *testing.T) {
+	t.Run("ResetSuccess", func(t *testing.T) {
+		res := initializeResponse(false, false, false, "")
+		err := CheckResponse(res)
+		if err != nil {
+			t.Errorf("Expected no errors, got %v", err)
+		}
+	})
+
+	t.Run("ResetError OS Rollback Impossible", func(t *testing.T) {
```
samribeiro, line 32: ResetError OS Rollback Unsupported

gnoi/reset/client_test.go
```diff
@@ -0,0 +1,70 @@
...
+		res := initializeResponse(true, false, false, "")
+		err := CheckResponse(res)
+		if err == nil {
+			t.Error("Expected OS unsupported error, no error returned")
+		}
+	})
+	t.Run("ResetError Zero Fill Impossible", func(t *testing.T) {
```
samribeiro, line 39: ResetError Zero Fill Unsupported

gnoi/reset/client_test.go
Jared-Mullin: Changed, thanks!

gnoi/reset/client_test.go
Jared-Mullin: Changed, thanks!

gnoi/reset/client.go
Jared-Mullin: I agree, I think I've written it better now

gnoi/reset/client_test.go
```diff
@@ -0,0 +1,70 @@
...
+package reset
+
+import (
+	"testing"
+
+	"github.com/google/gnxi/gnoi/reset/pb"
+)
+
+func TestStart(t *testing.T) {
+	t.Run("ResetSuccess", func(t *testing.T) {
```
samribeiro, line 24: Would this be more readable with:

```
// in client.go:
type ResetError struct{
    Msg string
}

func (re *ResetError)Error() string{
	return re.Msg
}

// in here:

	tests := []struct {
		response *pb.StartResponse{}
		want *ResetError
	}{
		{
			response: &pb.StartResponse{Response: &pb.StartResponse_ResetError{&pb.ResetError{
			           FactoryOsUnsupported: true,
			           ZeroFillUnsupported:  false,
			           Other:                false,
			           Detail:               "",
                         },},},
			want: errors.NewError("Factory OS Rollback Unsupported"),
		},
...
	}
	for _, test := range tests {
		got := CheckResponse(test.response)
                 diff := cmp.Diff(test.want, got)
		if diff != "" {
			t.Errorf("CheckResponse(%s): (-want +got):\n%s", test.response, diff)
		}
	}
}
```

gnoi/reset/client.go
```diff
@@ -0,0 +1,71 @@
...
+		resErr := res.GetResetError()
+		if resErr.FactoryOsUnsupported {
+			log.Println("Factory OS Rollback Unsupported")
```
samribeiro: Since we are already returning the error, lets leave the calling binary handle all the logging here.

samribeiro: and in other locations in this file.

samribeiro: in summary: remove logging =)

Jared-Mullin: Sure, no problem

gnoi/reset/client_test.go
Jared-Mullin: I think this is resolved in new commit, I kept the initializeResponse function but I can remove it if you prefer, thank you

gnoi/reset/client.go
```diff
@@ -0,0 +1,80 @@
...
+	"github.com/google/gnxi/gnoi/reset/pb"
+	"google.golang.org/grpc"
+)
+
+// Client handles requesting a Factory Reset.
+type Client struct {
+	client pb.FactoryResetClient
+}
+
+type ResetError struct {
```
ericm, line 32: Remember to comment

gnoi/reset/client.go
```diff
@@ -0,0 +1,80 @@
...
+	Msgs []string
+}
+
+func (re *ResetError) Error() string {
```
ericm, line 36: Probably comment here as well

gnoi/reset/client.go
Jared-Mullin: Fixed thank you :D

ericm: :)

gnoi/reset/client.go
```diff
@@ -0,0 +1,82 @@
...
+	"github.com/google/gnxi/gnoi/reset/pb"
+	"google.golang.org/grpc"
+)
+
+// Client handles requesting a Factory Reset.
+type Client struct {
+	client pb.FactoryResetClient
+}
+
+// ResetError allows multiple error messages
```
samribeiro, line 32: needs to be full sentence with a period.

gnoi/reset/client.go
```diff
@@ -0,0 +1,82 @@
...
+}
+
+// ResetTarget invokes gRPC start service on the server.
+func (c *Client) ResetTarget(ctx context.Context, zeroFill, rollbackOS bool) error {
+	out, err := c.client.Start(ctx, &pb.StartRequest{
+		FactoryOs: rollbackOS,
+		ZeroFill:  zeroFill,
+	})
+	if err != nil {
+		log.Println("Error calling Start service:", err)
```
samribeiro, line 54: we can also drop the log here and do it in the main package

gnoi/reset/client_test.go
```diff
@@ -0,0 +1,82 @@
...
+	"github.com/google/gnxi/gnoi/reset/pb"
+)
+
+func TestCheckResponse(t *testing.T) {
+	tests := []struct {
+		response *pb.StartResponse
+		want     *ResetError
+	}{
+		{
+			response: initializeResponse(true, false, false, ""),
```
samribeiro, line 31: just add directly here the following, it is easier to read:

```
response: &pb.StartResponse{&pb.StartResponse_ResetError{&pb.ResetError{
			FactoryOsUnsupported: true,
			ZeroFillUnsupported:  false,
			Other:                false,
			Detail:               "",
		},},},
```

Jared-Mullin: Sure, no problem, will change

gnoi/reset/client_test.go
```diff
@@ -0,0 +1,82 @@
...
+import (
+	"testing"
+
+	"github.com/google/go-cmp/cmp"
+
+	"github.com/google/gnxi/gnoi/reset/pb"
+)
+
+func TestCheckResponse(t *testing.T) {
+	tests := []struct {
```
samribeiro, line 26: missing:

```
response: &pb.StartResponse{&pb. StartResponse_ResetSuccess{&pb.Success{}}},
want: nil,
```

Jared-Mullin: I've included this in the tests now, I ran into some segmentation fail issues with cmp.Diff trying to compare nil versus nil, so I've used a continue in the loop to resolve this

samribeiro: if you return *ResetError here I think you wont have the issue:

func CheckResponse(res *pb.StartResponse) *ResetError {
}

would you try?

Jared-Mullin: That worked, thank you!

PR #103, opened 2020-07-01
Moved reusable code from gnoi_cert to utils

ericm:
These seem to be reusable for every client so I moved them to `utils`


ericm:
Got it. Thanks!

PR #104, opened 2020-07-02
gNOI Reset Client Binary and README

Jared-Mullin:
Refers to Issue #97
Added README.md and gnoi_reset client binary (based on gNMI binaries)

### DIFF
--- a/gnoi_reset/README.md
+++ b/gnoi_reset/README.md
@@ -1,0 +1,24 @@
+# gNOI Factory Reset Client
+
+A simple shell binary that performs Factory Reset client operations against a gNOI target.
+
+## gNOI Factory Reset Options
+*   `-rollback_os` will attempt to roll back the OS to the factory version and reset all certificates on the target. The target will then enter bootstrapping mode
+*   `-zero_fill` will attempt to zero fill the device’s persistent storage.
+
+## Install
+```
+go get github.com/google/gnxi/gnoi_reset
+go install github.com/google/gnxi/gnoi_reset
+```
+
+## Run 
+```
+gnoi_reset \
+-target_addr localhost:9399 \
+-rollback_os \
+-zero_fill \
+-key client.key \
+-cert client.crt \
+-ca ca.crt
+```
--- a/gnoi_reset/README.md
+++ b/gnoi_reset/README.md
@@ -1,9 +1,9 @@
 # gNOI Factory Reset Client
 
-A simple shell binary that performs Factory Reset client operations against a gNOI target.
+A simple shell binary that performs Factory Reset client operations against a gNOI target. The target will then enter bootstrapping mode.
 
 ## gNOI Factory Reset Options
-*   `-rollback_os` will attempt to roll back the OS to the factory version and reset all certificates on the target. The target will then enter bootstrapping mode
+*   `-rollback_os` will attempt to roll back the OS to the factory version and reset all certificates on the target.
 *   `-zero_fill` will attempt to zero fill the device’s persistent storage.
 
 ## Install
--- a/gnoi_reset/gnoi_reset.go
+++ b/gnoi_reset/gnoi_reset.go
@@ -1,0 +1,61 @@
+/* Copyright 2020 Google Inc.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    https://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+// Binary implements a gNOI Reset.
+
+package main
+
+import (
+	"context"
+	"flag"
+	"time"
+
+	"google.golang.org/grpc"
+
+	log "github.com/golang/glog"
+
+	"github.com/google/gnxi/gnoi/reset"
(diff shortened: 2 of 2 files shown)


gnoi_reset/gnoi_reset.go
```diff
@@ -0,0 +1,61 @@
...
+		log.Exitf("Dialing to %s failed: %v", *targetAddr, err)
+	}
+	defer conn.Close()
+
+	client := reset.NewClient(conn)
+
+	ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
+	defer cancel()
+
+	err = client.ResetTarget(ctx, *rollbackOs, *zeroFill)
```
loriscn12, line 55: I think the following will reduce this by 1 line:

if err := client.ResetTarget(ctx, *rollbackOs, *zeroFill); err != nil {
...
...

Jared-Mullin: I agree

gnoi_reset/gnoi_reset.go
```diff
@@ -0,0 +1,60 @@
...
+	"google.golang.org/grpc"
+
+	log "github.com/golang/glog"
+
+	"github.com/google/gnxi/gnoi/reset"
+	"github.com/google/gnxi/utils/credentials"
+)
+
+var (
+	targetAddr = flag.String("target_addr", "localhost:9399", "The target address in the format of host:port")
```
brianokelly1, line 34: Probably best not to provide defaults here.

gnoi_reset/gnoi_reset.go
```diff
@@ -0,0 +1,60 @@
...
+	targetName = flag.String("target_name", "hostname.com", "The target name used to verify the hostname returned by TLS handshake")
+	rollbackOs = flag.Bool("rollback_os", false, "Indicate if target should attempt to revert to factory os")
+	zeroFill   = flag.Bool("zero_fill", false, "Indicate if target should attempt to overwrite persistent storage with zeroes")
+)
+
+func main() {
+	flag.Parse()
+
+	opts := credentials.ClientCredentials(*targetName)
```
brianokelly1, line 43: I would check that targetName is set, if it is not, then print usage and exit.

gnoi_reset/gnoi_reset.go
```diff
@@ -0,0 +1,60 @@
...
+	conn, err := grpc.Dial(*targetAddr, opts...)
+	if err != nil {
+		log.Exitf("Dialing to %s failed: %v", *targetAddr, err)
+	}
+	defer conn.Close()
+
+	client := reset.NewClient(conn)
+
+	ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
```
brianokelly1, line 52: Time should be a flag also or at the minimum a const.

Jared-Mullin: Sure, changing to flag with a default of 10 seconds

gnoi_reset/gnoi_reset.go
Jared-Mullin: Defaults are now removed

gnoi_reset/gnoi_reset.go
Jared-Mullin: I've changed it so that both targetAddr and targetName are checked, if they are not set it prints usage and exits specifying those specific flags

gnoi_reset/gnoi_reset.go
Jared-Mullin: Added a timeout flag now

gnoi_reset/README.md
```diff
@@ -0,0 +1,24 @@
...
+## Install
+```
+go get github.com/google/gnxi/gnoi_reset
+go install github.com/google/gnxi/gnoi_reset
+```
+
+## Run 
+```
+gnoi_reset \
+-target_addr localhost:9399 \
```
brianokelly1, line 18: Also add target_name to the example here.

gnoi_reset/gnoi_reset.go
```diff
@@ -0,0 +1,66 @@
...
+	"github.com/google/gnxi/gnoi/reset"
+	"github.com/google/gnxi/utils/credentials"
+)
+
+var (
+	targetAddr = flag.String("target_addr", "", "The target address in the format of host:port")
+	targetName = flag.String("target_name", "", "The target name used to verify the hostname returned by TLS handshake")
+	rollbackOs = flag.Bool("rollback_os", false, "Target must revert to factory OS")
+	zeroFill   = flag.Bool("zero_fill", false, "Target must overwrite persistent storage with zeroes")
+	timeOut    = flag.Duration("time_out", 10*time.Second, "Timeout for ResetTarget operation, 10 seconds by default")
```
brianokelly1, line 38: nit: Best to give more descriptive name in case other timeouts are added later. Maybe: "reset_target_timeout".

nit: I would also include an example of what units are required. flag.Duration takes anything parsable by flag.ParseDuration so for example the user could use:

"Timeout for ResetTarget operation, 10 seconds by default e.g. 10s"

gnoi_reset/gnoi_reset.go
```diff
@@ -0,0 +1,66 @@
...
+	}
+	defer conn.Close()
+
+	client := reset.NewClient(conn)
+
+	ctx, cancel := context.WithTimeout(context.Background(), *timeOut)
+	defer cancel()
+
+	if err = client.ResetTarget(ctx, *rollbackOs, *zeroFill); err != nil {
+		log.Errorf("Error Resetting Target: %v", err)
```
brianokelly1, line 62: I think we might want log.Exitf here as we would also want os.Exit(1) called to indicate to a caller that the operation failed.

PR #105, opened 2020-07-02
Added gNOI reset to target binary and implemented cert resetting

ericm:
I pulled in the certServer to the module so it can interface with the 2 RPCs it needs


gnoi/reset/server.go
```diff
@@ -18,10 +18,17 @@ package reset
 import (
 	"context"
 
+	certPB "github.com/google/gnxi/gnoi/cert/pb"
 	"github.com/google/gnxi/gnoi/reset/pb"
 	"google.golang.org/grpc"
 )
 
+// CertServerInterface used to pass in cert server for factory reset
```
loriscn12, line 26: end with period.

gnoi/reset/server.go
```diff
@@ -18,10 +18,17 @@ package reset
 import (
 	"context"
 
+	certPB "github.com/google/gnxi/gnoi/cert/pb"
```
loriscn12, line 21: s/certPB/cpb

gnoi/reset/server.go
```diff
@@ -54,7 +62,27 @@ func (s *Server) Start(ctx context.Context, req *pb.StartRequest) (*pb.StartResp
...
+
+	certs := []string{}
+	for _, c := range response.CertificateInfo {
+		certs = append(certs, c.CertificateId)
+	}
+
+	_, err = s.certServer.RevokeCertificates(context.Background(), &certPB.RevokeCertificatesRequest{CertificateId: certs})
+	if err != nil {
+		return err
+	}
```
loriscn12, line 86:
```suggestion
	if _, err := s.certServer.RevokeCertificates(context.Background(), &certPB.RevokeCertificatesRequest{CertificateId: certs}); err != nil {
		return err
	}
```

ericm:
Should be fixed now :)

ericm:
#106

gnoi/reset/server_test.go
```diff
@@ -60,8 +72,9 @@ func makeTests() []test {
 
 func TestStart(t *testing.T) {
 	tests := makeTests()
+	mock := &certServerMock{}
```
brianokelly1, line 75: Use a more descriptive name here. You could end up with lots of different things mocked and it would make it easier to understand which is which.

gnoi/reset/server.go
```diff
@@ -54,7 +62,26 @@ func (s *Server) Start(ctx context.Context, req *pb.StartRequest) (*pb.StartResp
 		return &pb.StartResponse{Response: &pb.StartResponse_ResetError{ResetError: resetError}}, nil
 	}
 
-	// TODO: Reset the Target.
+	s.reset()
 
 	return &pb.StartResponse{Response: &pb.StartResponse_ResetSuccess{}}, nil
 }
+
+// reset the target device. Clears certs and wipes OS's
```
brianokelly1, line 70: nit: Reset

gnoi/reset/server.go
```diff
@@ -54,7 +62,26 @@ func (s *Server) Start(ctx context.Context, req *pb.StartRequest) (*pb.StartResp
 		return &pb.StartResponse{Response: &pb.StartResponse_ResetError{ResetError: resetError}}, nil
 	}
 
-	// TODO: Reset the Target.
+	s.reset()
 
 	return &pb.StartResponse{Response: &pb.StartResponse_ResetSuccess{}}, nil
```
brianokelly1, line 67: It's not clear to me why they ever return an error here since they indicate failures in the StartResponse. We might consider in the future changing this. Just food for thought!

gnoi/reset/server_test.go
```diff
@@ -29,6 +30,17 @@ type test struct {
...
 }
 
+type certServerMock struct {
+	CertServerInterface
+}
+
+func (m *certServerMock) GetCertificates(ctx context.Context, req *cpb.GetCertificatesRequest) (*cpb.GetCertificatesResponse, error) {
+	return &cpb.GetCertificatesResponse{CertificateInfo: []*cpb.CertificateInfo{}}, nil
+}
+func (m *certServerMock) RevokeCertificates(ctx context.Context, req *cpb.RevokeCertificatesRequest) (*cpb.RevokeCertificatesResponse, error) {
```
brianokelly1, line 40: nit: add blank line between functions.

gnoi/reset/server.go
ericm: I think its the way the grpc handlers are setup. Not sure if theres a way to easily change the signature of them

samribeiro: Such that we can send gRPC specific errors as oposed to service specific errors. We are doing the latter.
https://godoc.org/google.golang.org/grpc/status#Error

gnoi_target/README.md
```diff
@@ -8,6 +8,11 @@ Management service, supporting bootstrapping mode.
 This service provides a set of RPCs to Install, Rotate & Revoke Certificates and
 CA Bundles in a Target. See [gNOI Cert proto definition](https://github.com/openconfig/gnoi/blob/master/cert/cert.proto) for more.
 
+## Factory Reset service
+
+This service provides an RPC to Start a factory reset of the device. This includes
+resetting all certificates on the device. See [gNOI Cert proto definition](https://github.com/openconfig/gnoi/blob/master/factory_reset/reset.proto) for more.
```
samribeiro, line 14: resetting all certificates on the device and setting it to bootstrapped mode.

gnoi_target/gnoi_target.go
```diff
@@ -51,18 +56,27 @@ func serve() {
 	}
 }
 
-// notify can be called with the number of certs and ca certs installed. It will
+// notifyReset will be called when the factory reset service requires the server
+// to be restarted.
+func notifyReset() {
+	var err error
+	if gNOIServer, err = gnoi.NewServer(nil, nil, resetSettings); err != nil {
```
samribeiro, line 63: things that you need to consider here:
1) you need to close the existing connections and server being handled in serve().
2) this is the place to introduce the delay, you can even send a log message.
3) you need to call notifyCerts at the end again to restart the server.

ericm: I think doing number 3 will be enough for 1 since it gracefully shuts down the server

ericm:
Pushed some fixes :)

ericm:
Should be fine now :)

gnoi_target/gnoi_target.go
```diff
@@ -51,42 +56,61 @@ func serve() {
...
 		log.Fatal("Failed to create gNOI Server:", err)
 	}
-
 	// Registers a caller for whenever the number of installed certificates changes.
-	gNOIServer.RegisterNotifier(notify)
-	notify(0, 0) // Trigger bootstraping mode.
-	select {}    // Loop forever.
+	gNOIServer.RegisterCertNotifier(notifyCerts)
+	bootstrapping = false
+	notifyCerts(0, 0) // Trigger bootstraping mode.
```
loriscn12, line 101: s/Trigger/Triggers

gnoi/reset/server.go
brianokelly1: Ack fair enough. It's more a thing of if we never set error, then its pretty useless but if we are following a particular design pattern then ok.

Issue #107, opened 2020-07-02
Implement OS service manager for mock target


ericm:
This will hold the state for what OS is running, what's the factory OS and all other available OS's.

PR #109, opened 2020-07-02
Added manager for OS service

ericm:
Related issue: #107 
The manager will hold the state of the OS's installed for the OS service.


gnoi/os/manager.go
```diff
@@ -0,0 +1,100 @@
...
+		return nil
+	}
+	if supported := m.supportedVersions[o.Version]; !supported {
+		return fmt.Errorf("INCOMPATIBLE")
+	}
+	m.osMap[o.Version] = o
+	return nil
+}
+
+// Rollback to factory OS. Wipes all other OS's.
```
loriscn12, line 90: Rollbacks ... and wipes...

gnoi/os/manager.go
```diff
@@ -0,0 +1,100 @@
...
+
+// SetRunning OS to new version. OS must exist in manager.
+func (m *Manager) SetRunning(version string) error {
+	if o, ok := m.osMap[version]; ok && o != nil {
+		m.running = o
+	}
+	return fmt.Errorf("NON_EXISTENT_VERSION")
+}
+
+// Install an OS. Must be fully transfered and verified beforehand.
```
loriscn12, line 78: Installs

gnoi/os/manager.go
```diff
@@ -0,0 +1,100 @@
...
+func (m *Manager) Install(o *OS) error {
+	if _, ok := m.osMap[o.Version]; ok {
+		return nil
+	}
+	if supported := m.supportedVersions[o.Version]; !supported {
+		return fmt.Errorf("INCOMPATIBLE")
+	}
+	m.osMap[o.Version] = o
+	return nil
```
loriscn12, line 87:
```suggestion
	if supported := m.supportedVersions[o.Version]; !supported {
		return fmt.Errorf("INCOMPATIBLE")
	}
	if _, ok := m.osMap[o.Version]; !ok {
		m.osMap[o.Version] = o
	}
	return nil
```

gnoi/os/manager.go
```diff
@@ -0,0 +1,100 @@
...
+
+// IsRunning the version of OS specified.
+func (m *Manager) IsRunning(version string) bool {
+	if m.running != nil {
+		return m.running.Version == version
+	}
+	return false
+}
+
+// SetRunning OS to new version. OS must exist in manager.
```
loriscn12, line 70: Sets the running OS...

gnoi/os/manager.go
```diff
@@ -0,0 +1,100 @@
...
+func (m *Manager) SetRunning(version string) error {
+	if o, ok := m.osMap[version]; ok && o != nil {
```
loriscn12, line 72: can the entry be nil when ok?

ericm:
The transfer state is all going to be handled in the OS Server. This will just manage the OS's after the fact

gnoi/os/manager.go
```diff
@@ -0,0 +1,100 @@
...
+func NewManager(supported map[string]bool, factoryOs *OS) *Manager {
+	return &Manager{
+		standbyState:      unsupported,
+		supportedVersions: supported,
+		factory:           factoryOs,
+		osMap:             map[string]*OS{factoryOs.Version: factoryOs},
+	}
+}
+
+// IsRunning the version of OS specified.
```
loriscn12, line 62: drop it or make it more descriptive

gnoi/os/manager.go
ericm: It can be but in normal circumstances I suppose it's redundant

gnoi/os/manager.go
ericm: Does the first word of the comment not have to be the method name?

loriscn12: Yep leave that , you can have:

Install installs an OS ...

gnoi/os/manager.go
```diff
@@ -0,0 +1,99 @@
...
+// OS represents the data for a singular operating system.
+type OS struct {
+	Name,
+	Hash,
+	Version string
+	BuildDate time.Time
+	Data      *bytes.Buffer
+}
+
+// Manager for storing data on OS's.
```
samribeiro, line 33: I like how you have set this up has if we are running real OSs in the Target =)

I think we can simplify here a bit (this is what I think, I may be wrong, please point out if so):
1) Manager only really needs: runningOS, factoryOS and []installeOSs string. We dont do anything with all the other data.
2) The hash validation is only done at the reception of the binary data, so we dont need to store the hashes;
3) Manager should support all versions, we can signal an unsupported version in the OS file directly by using a cookie, example:

OS file:
hash (for everything after this line)\
cookie (just a unique id string) \
version (string) \
random padding \
<EOF>

ericm: So for now just keep OS here and later move it?

ericm:
I simplified it according to your suggestions :)

gnoi/os/manager.go
```diff
@@ -0,0 +1,54 @@
...
+	return version == m.runningVersion
+}
+
+// SetRunning sets the running OS to the version specified.
+func (m *Manager) SetRunning(version string) error {
+	if _, ok := m.osMap[version]; ok {
+		m.runningVersion = version
+		return nil
+	}
+	return fmt.Errorf("NON_EXISTENT_VERSION")
```
samribeiro, line 48: we may need to revisit this after but good for now.

PR #117, opened 2020-07-06
Added tests for os manager

### LINKED ISSUE #107: Implement OS service manager for mock target

ericm:
Related issue: Closes #107

### DIFF
--- a/gnoi/os/manager_test.go
+++ b/gnoi/os/manager_test.go
@@ -53,7 +53,7 @@ func TestSetRunning(t *testing.T) {
 		t.Run(test.name, func(t *testing.T) {
 			manager := NewManager("new")
 			err := manager.SetRunning(test.version)
-			if (err == nil) != test.wantsErr {
+			if (err == nil) == test.wantsErr {
 				t.Errorf("Error running OS: Error is %w", err)
 			}
 		})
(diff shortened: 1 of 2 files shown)


gnoi/os/manager_test.go
```diff
@@ -0,0 +1,61 @@
...
+			t.Error("OS not running after being set running")
+		}
+	})
+	t.Run("Is not running", func(t *testing.T) {
+		manager := NewManager("new")
+		if manager.IsRunning("new") {
+			t.Error("OS running but not set running")
+		}
+	})
+}
```
loriscn12, line 34: again not tested but, is there any way you could make the following work? I am not specifically against repetitive code especially if it's clean but I guess in this case this can be shortened. Let me know otherwise:

```suggestion
func TestIsRunning(t *testing.T) {
       mc := NewManager("new")
       tests = tests := []struct {
		name string
		mgrClient *Manager
		runningVersion string
		want bool
	}{
		{
			name: "false"
			mgrClient: mc,
			rVer: "new"
			want: false,
		},
	        {
	                name: "true"
			mgrClient: mc,
			rVer: nil,
			want: true,
	        },	
	}
	for _, test := range tests{
		t.Run(t.name, func(t *testing.T) {
			manager.runningVersion = t.rVer
			got := manager.IsRunning("new")
			if  got != t.want {
				t.Errorf("want %s got %v", want, got)
			}
		})
	}
}
```

PS. I am not sure you can use one client for both tests, if not just init a new one every time np.

ericm:
Made the tests more concise :)

PR #118, opened 2020-07-06
Added gNOI Client Activate RPC caller

### LINKED ISSUE #113: Implement OS Activate RPC for mock target

ericm:
Related issue: Closes #113 
Implemented Activate RPC


gnoi/os/client.go
```diff
@@ -0,0 +1,81 @@
...
+
+// Enum representing possible ActivateErrorTypes.
+const (
+	ActivateUnspecified        ActivateErrorType = "UNSPECIFIED"
+	ActivateNonExistentVersion                   = "NON_EXISTENT_VERSION"
+)
+
+// ActivateError represents an error returned by the Activate RPC.
+type ActivateError struct {
+	error
```
samribeiro, line 42: is embeding of error required?

gnoi/os/client.go
```diff
@@ -0,0 +1,81 @@
...
+	out, err := c.client.Activate(ctx, &pb.ActivateRequest{Version: version})
+	if err != nil {
+		return err
+	}
+	switch out.Response.(type) {
+	case *pb.ActivateResponse_ActivateOk:
+		return nil
+	case *pb.ActivateResponse_ActivateError:
+		res := out.GetActivateError()
+		errType := ActivateErrorType(res.GetType().String())
```
samribeiro, line 67: these already exist in the proto generated code?
https://github.com/google/gnxi/blob/master/gnoi/os/pb/os.pb.go#L118

gnoi/os/client.go
```diff
@@ -0,0 +1,81 @@
...
+		switch errType {
+		case ActivateUnspecified:
+			fallthrough
+		case ActivateNonExistentVersion:
+			return &ActivateError{
```
samribeiro, line 72: the detail is only used for unspecified errors. All the others need not use the detail field.

gnoi/os/client.go
ericm: I suppose I defined them in the os package so you wont have to import the proto generated code to do a type check in the main binary. As well as that these types contain strings so all you'd have to do is print it to the end user :) I can add another type for other error if you want

gnoi/os/client_test.go
```diff
@@ -0,0 +1,92 @@
...
+
+func TestActivate(t *testing.T) {
+	activateTests := generateActivateTests()
+	for _, test := range activateTests {
+		t.Run(test.name, func(t *testing.T) {
+			client := Client{client: test.client}
+			got := client.Activate(context.Background(), "version")
+			if got == nil && test.want == nil {
+				return
+			}
```
loriscn12, line 86: you want to test negative results, do not return but check if it fails and print something like:

t.Errorf("want nil got non-nil")

ericm:
Due to the fact that nil doesnt equal nil in this case i couldnt compare the nil error case against each other

gnoi/os/client_test.go
```diff
@@ -0,0 +1,93 @@
...
+		t.Run(test.name, func(t *testing.T) {
+			client := Client{client: test.client}
+			got := client.Activate(context.Background(), "version")
+			if got == nil || test.want == nil {
+				if (got == nil && test.want != nil) || (got != nil && test.want == nil) {
+					t.Errorf("Wanted: %v, Got: %v", test.want, got)
+				}
+			} else if err := got.(*ActivateError); err.ErrType != test.want.ErrType || err.Detail != test.want.Detail {
+				t.Errorf("Wanted: ActivateError(%v), Got: ActivateError(%v)", *test.want, *err)
+			}
```
loriscn12, line 90: I am not 100% sure cause I haven't tested it but I think you could simplify it al follow:

```
err := got.(*ActivateError)
if want != nil && (err.ErrType != test.want.ErrType || err.Detail != test.want.Detail) {
    t.Errorf("want: ActivateError(%v), got: ActivateError(%v)", *test.want, *err)
  }
} else if got != nil {
    t.Errorf("want <nil>, got: %v", got)
  }
}
```

ericm: Afraid not. Issue is the nil checks.
```

--- FAIL: TestActivate (0.00s)
    --- FAIL: TestActivate/Success (0.00s)
panic: interface conversion: error is nil, not *os.ActivateError [recovered]
	panic: interface conversion: error is nil, not *os.ActivateError
```

loriscn12: I tested the following:

			if test.want != nil {
				err := got.(*ActivateError)
				if err.ErrType != test.want.ErrType || err.Detail != test.want.Detail {
					t.Errorf("want: ActivateError(%v), got: ActivateError(%v)", *test.want, *err)
				}
			} else if got != nil {
				t.Errorf("want <nil>, got: %v", got)
			}

I think this works, but double check, if not and no other way to simplify just keep it as is ;)

gnoi/os/client.go
samribeiro: oh I see what your plan is. I think you can actually keep the pb scoped to this file only. All the client here needs to do is to abstract the pb stuff from the main. The main only calls the actions and receives a (string) error. Why would it need more than that? =). This way you can simplify the error handling and avoid having to redefine new consts in this file every time the proto gets augmented.

```
	switch out.Response.(type) {
	case *pb.ActivateResponse_ActivateOk:
		return nil
	case *pb.ActivateResponse_ActivateError:
		res := out.GetActivateError()
		switch res.GetType() {
		case pb.ActivateError_UNSPECIFIED:
			return fmt.Errorf("Unspecified ActivateError: %s", res.GetDetail())
		case pb.ActivateError_NON_EXISTENT_VERSION:
			return fmt.Errorf("Non existent version: %s", version)
		default:
			return fmt.Errorf("Unknown ActivateError: %s", res.GetType())
		}
	}
```

ericm: Thats true. Thanks!

ericm:
Should be simplified now :)

PR #122, opened 2020-07-07
Fixing manger_test.go case.

loriscn12:
Context: https://github.com/golang/go/issues/35472

### DIFF
--- a/gnoi/os/manager_test.go
+++ b/gnoi/os/manager_test.go
@@ -54,7 +54,7 @@ func TestSetRunning(t *testing.T) {
 			manager := NewManager("new")
 			err := manager.SetRunning(test.version)
 			if (err == nil) == test.wantsErr {
-				t.Errorf("Error running OS: Error is %w", err)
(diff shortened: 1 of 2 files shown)


gnoi/os/manager_test.go
```diff
@@ -54,7 +54,7 @@ func TestSetRunning(t *testing.T) {
 			manager := NewManager("new")
 			err := manager.SetRunning(test.version)
 			if (err == nil) == test.wantsErr {
-				t.Errorf("Error running OS: Error is %w", err)
+				t.Errorf("Error running OS: Error is %v", err)
```
ericm, line 57: What version of go did you run into issues with `%w`?

ericm: It was introduced in 1.13

ericm: https://blog.golang.org/go1.13-errors

loriscn12: my bad, I used my personal laptop and did not check the go version, nice you spotted this :)

PR #127, opened 2020-07-08
Added Verify RPC call for OS client

### LINKED ISSUE #125: Implement OS Verify RPC for Client

ericm:
Related Issue: Closes #125

### DIFF
--- a/gnoi/os/client.go
+++ b/gnoi/os/client.go
@@ -55,3 +55,17 @@ func (c *Client) Activate(ctx context.Context, version string) error {
 	}
 	return nil
 }
+
+// Verify invokes the Verify RPC for the OS service.
+// It returns the running version, activation fail message and error
+func (c *Client) Verify(ctx context.Context) (string, string, error) {
+	out, err := c.client.Verify(ctx, &pb.VerifyRequest{})
+	if err != nil {
+		return "", "", err
+	}
+	outVer := out.GetVersion()
+	if fail := out.GetActivationFailMessage(); fail != "" {
+		return outVer, fail, nil
+	}
+	return outVer, "", nil
+}
--- a/gnoi/os/client.go
+++ b/gnoi/os/client.go
@@ -1,0 +1,57 @@
+/* Copyright 2020 Google Inc.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    https://www.apache.org/licenses/LICENSE-2.0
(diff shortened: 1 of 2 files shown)


ericm:
I decided to return activation fail separate to error since it indicates a previous error rather than a current one

gnoi/os/client.go
```diff
@@ -55,3 +55,16 @@ func (c *Client) Activate(ctx context.Context, version string) error {
 	}
 	return nil
 }
+
+// Verify invokes the Verify RPC for the OS service and it returns the running version and activation fail message or an error.
+func (c *Client) Verify(ctx context.Context) (string, string, error) {
```
samribeiro, line 60: One of the few cases it is worth naming the return variables.

```
func (c *Client) Verify(ctx context.Context) (version, failMsg string, err error) {
        var out *pb.VerifyResponse
	if out, err = c.client.Verify(ctx, &pb.VerifyRequest{}); err != nil {
		return
	}
	version = out.GetVersion()
        failMsg = out.GetActivationFailMessage()
	return
}
```

ericm: Oh that's a great idea

ericm:
Should be done :)

gnoi/os/client_test.go
```diff
@@ -36,10 +38,20 @@ type activateTest struct {
...
+	failMessage string
+}
+
 func (c *mockClient) Activate(ctx context.Context, in *pb.ActivateRequest, opts ...grpc.CallOption) (*pb.ActivateResponse, error) {
 	return c.activate(ctx, in, opts...)
 }
 
+func (c *mockClient) Verify(ctx context.Context, in *pb.VerifyRequest, opts ...grpc.CallOption) (*pb.VerifyResponse, error) {
+	return c.verify(ctx, in, opts...)
+}
```
loriscn12, line 53: I like the approach ;)

gnoi/os/client_test.go
```diff
@@ -36,10 +38,20 @@ type activateTest struct {
 	wantErr bool
 }
 
+type verifyTest struct {
+	name,
+	runningVersion,
+	failMessage string
+}
```
samribeiro, line 45: you are only using this once, perhaps making it inline downthere is better and more readable?

```
tests := []struct {
        name,
        runningVersion,
        failMessage string
}{
        {"Is Running", "version", ""},
        {"Previous Activation Fail", "version", "Activation fail"},
}
```

ericm:
Done :)

PR #131, opened 2020-07-09
Added Proto Logging for gNOI OS, Cert and Reset

### LINKED ISSUE #120: Add a debug flag that enables dumping the RPC proto messages to the log
use this:
https://github.com/google/gnxi/blob/a11b4f247cf18f30dfaba94999845b18294f0bb5/utils/utils.go#L17

Jared-Mullin:
Closes #120 
Utils has a new LogProto function and a new log_proto flag

### DIFF
--- a/gnoi/cert/client.go
+++ b/gnoi/cert/client.go
@@ -65,12 +65,12 @@ func (c *Client) Rotate(ctx context.Context, certID string, minKeySize uint32, p
 	if err = stream.Send(request); err != nil {
 		return fmt.Errorf("failed to send GenerateCSRRequest: %v", err)
 	}
-	var req *pb.RotateCertificateResponse
-	if req, err = stream.Recv(); err != nil {
+	var response *pb.RotateCertificateResponse
+	if response, err = stream.Recv(); err != nil {
 		return fmt.Errorf("failed to receive RotateCertificateResponse: %v", err)
 	}
-	utils.LogProto(req)
-	genCSR := req.GetGeneratedCsr()
+	utils.LogProto(response)
+	genCSR := response.GetGeneratedCsr()
 	if genCSR == nil || genCSR.Csr == nil {
 		return fmt.Errorf("expected GenerateCSRRequest, got something else")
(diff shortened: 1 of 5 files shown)


Jared-Mullin:
Just adding Loris' comment from before about Rotate:
This function is >100 lines and as I understand it, it implements three well defined operations in it (Generate, Load, Finalize). For the sake of readability, testability etc.. is there any way you could split this and have three separate functions all called in here? This is just my opinion honestly, I'll it to you guys and @samribeiro to decide :)

utils/utils.go
```diff
@@ -21,3 +23,10 @@ func PrintProto(m proto.Message) {
 	}
 	fmt.Println(proto.MarshalTextString(m))
 }
+
+// LogProto prints a Proto message only if the log_proto flag is true.
+func LogProto(m proto.Message) {
```
samribeiro, line 28: allow using pretty here too. You can refactor the PrintProto function:

```
func FormatProto(m proto.Message) string {
        if *usePretty {
		return pretty.Sprint(m)
	}
	return proto.MarshalTextString(m)
}

func PrintProto(m proto.Message) {
      fmt.Println(FormatProto(m))
}

func LogProto(m proto.Message){
     if *logProto {
		log.Info(FormatProto(m))
     }
}
```

gnoi/cert/client.go
```diff
@@ -42,30 +43,33 @@ func (c *Client) Rotate(ctx context.Context, certID string, minKeySize uint32, p
 	if err != nil {
 		return fmt.Errorf("failed stream: %v", err)
 	}
-	if err = stream.Send(&pb.RotateCertificateRequest{
+	request := &pb.RotateCertificateRequest{
```
samribeiro, line 46: You started a good thing here =)
would you be so kind as to rename all request variables as requests and response messages as response? But only in the code you are touching here. Thanks!

Jared-Mullin: No problem, will do

PR #132, opened 2020-07-09
Added OS Client Install RPC

### LINKED ISSUE #126: Implement OS Install RPC for Client

ericm:
Related issue: Closes #126

### DIFF
--- a/gnoi/os/client.go
+++ b/gnoi/os/client.go
@@ -136,8 +136,12 @@ func (c *Client) validateInstallRequest(response *pb.InstallResponse) (progress 
 		}
 		err = fmt.Errorf("InstallError occured: %s", installErr.GetType().String())
 		return
+	case *pb.InstallResponse_TransferReady:
+		return
+	default:
+		err = fmt.Errorf("Unexpected response: %T(%v)", resp, resp)
+		return
 	}
-	return
 }
 
 func (c *Client) accumulateErrors(recvErrs chan error) error {
--- a/gnoi/os/client.go
+++ b/gnoi/os/client.go
@@ -46,6 +46,7 @@ func (c *Client) Install(ctx context.Context, imgPath, version string, printStat
 		return err
 	}
 	buffer := bytes.NewBuffer(file)
+	fileSize := uint64(buffer.Len())
 	install, err := c.client.Install(ctx)
 	defer install.CloseSend()
 	if err != nil {
@@ -88,7 +89,7 @@ func (c *Client) Install(ctx context.Context, imgPath, version string, printStat
 				return
 			}
 			if printStatus {
-				fmt.Printf("%d%% transferred\n", progress)
+				fmt.Printf("%d%% transferred\n", progress/fileSize)
 			}
 		}
 	}()
@@ -119,13 +120,13 @@ func (c *Client) Install(ctx context.Context, imgPath, version string, printStat
 	}
 }
 
-func (c *Client) validateInstallRequest(response *pb.InstallResponse) (progress uint32, validated bool, err error) {
+func (c *Client) validateInstallRequest(response *pb.InstallResponse) (progress uint64, validated bool, err error) {
 	switch resp := response.Response.(type) {
 	case *pb.InstallResponse_Validated:
 		validated = true
 		return
-	case *pb.InstallResponse_SyncProgress:
-		progress = resp.SyncProgress.GetPercentageTransferred()
+	case *pb.InstallResponse_TransferProgress:
+		progress = resp.TransferProgress.GetBytesReceived()
 		return
 	case *pb.InstallResponse_InstallError:
 		installErr := resp.InstallError
--- a/gnoi/os/client.go
+++ b/gnoi/os/client.go
@@ -179,7 +179,7 @@ func (c *Client) validateInstallRequest(response *pb.InstallResponse) (progress 
 	}
 }
 
-// accumulateErrors will deplete erros from a channel and consolidate them.
+// accumulateErrors will deplete errors from a channel and consolidate them.
 func (c *Client) accumulateErrors(recvErrs chan error) error {
 	err := <-recvErrs
 	if err == nil {
--- a/gnoi/os/client.go
+++ b/gnoi/os/client.go
@@ -22,6 +22,7 @@ import (
 	"os"
 	"time"
 
+	log "github.com/golang/glog"
 	"github.com/google/gnxi/gnoi/os/pb"
 	"github.com/google/gnxi/utils"
 	"google.golang.org/grpc"
@@ -91,6 +92,7 @@ func (c *Client) Install(ctx context.Context, imgPath, version string, printStat
 	utils.LogProto(transferResp)
 	switch resp := transferResp.Response.(type) {
 	case *pb.InstallResponse_Validated:
+		log.Info("OS already validated")
 		return nil
 	case *pb.InstallResponse_InstallError:
 		installErr := resp.InstallError
--- a/gnoi/os/client.go
+++ b/gnoi/os/client.go
@@ -83,14 +83,14 @@ func (c *Client) Install(ctx context.Context, imgPath, version string, printStat
 	if err != nil {
 		return err
 	}
-	_, validated, err := c.validateInstallRequest(transferResp)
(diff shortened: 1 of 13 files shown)


gnoi/os/client.go
```diff
@@ -28,11 +32,129 @@ type Client struct {
...
+			}
+			if validated {
+				recvValidated <- validated
+				return
+			}
+			if printStatus {
+				fmt.Printf("%d%% transferred\n", progress/fileSize)
+			}
+		}
+	}()
```
loriscn12, line 95: add a comment before this go func here.

gnoi/os/client.go
```diff
@@ -28,11 +32,129 @@ type Client struct {
...
+			return err
+		}
+		err = install.Send(&pb.InstallRequest{
+			Request: &pb.InstallRequest_TransferContent{TransferContent: b},
+		})
+		if err != nil {
+			return err
+		}
+	}
+
```
loriscn12, line 112: add a comment here too.

gnoi/os/client.go
```diff
@@ -28,11 +32,129 @@ type Client struct {
...
 
+const chunkSize = 5000000
+
 // NewClient returns a new OS service client.
 func NewClient(c *grpc.ClientConn) *Client {
 	return &Client{client: pb.NewOSClient(c)}
 }
 
+// Install invokes the Install RPC for the OS service.
+func (c *Client) Install(ctx context.Context, imgPath, version string, printStatus bool, validateTimeout time.Duration) error {
```
loriscn12, line 43: i think this looks nice, however don't assume future readers will be all familiar with the logic you have implemented, specifically because this function is very long, add comments explaining each section briefly but in a descriptive way. A good example would be:

// Check versioning using RPC X

// Transferring content asynchronously using goroutine

// Checking buffer while streaming content

// Conferming transfer

etc...

gnoi/os/client.go
```diff
@@ -28,11 +32,129 @@ type Client struct {
...
+		return
+	default:
+		err = fmt.Errorf("Unexpected response: %T(%v)", resp, resp)
+		return
+	}
+}
+
+func (c *Client) accumulateErrors(recvErrs chan error) error {
+	err := <-recvErrs
+	if err == nil {
```
loriscn12, line 149: you should be ok making this one line.

ericm: I can but it causes golint to trigger sincei ave to put the rest in an else statement.
```go
	if err := <-recvErrs; err == nil {
		return nil
	} else {
		for len(recvErrs) > 0 {
			err = fmt.Errorf("%s; %w", (<-recvErrs).Error(), err)
		}
		return err
	}
```
`if block ends with a return statement, so drop this else and outdent its block (move short variable declaration to its own line if necessary)go-lint`

ericm:
Done :)

gnoi/os/client.go
```diff
@@ -28,11 +32,139 @@ type Client struct {
...
+
 // NewClient returns a new OS service client.
 func NewClient(c *grpc.ClientConn) *Client {
 	return &Client{client: pb.NewOSClient(c)}
 }
 
+// Install invokes the Install RPC for the OS service.
+func (c *Client) Install(ctx context.Context, imgPath, version string, printStatus bool, validateTimeout time.Duration) error {
+	// Read file to buffer.
+	file, err := ioutil.ReadFile(imgPath)
```
samribeiro, line 45: I know this complicates things, but imagine the OS file is 1G (normal for some platforms). Would it be better to load chunks of it instead? =)

ericm: Oh i get you. Should be done :)

ericm:
I switched from reading to a buffer to using `file.ReadAt` each iteration to just load each chunk before sending :)

ericm:
Reformatted some of it :)

gnoi/os/client.go
```diff
@@ -27,11 +30,170 @@ import (
...
+		return err
+	}
+	defer fileClose()
+
+	// Create Install client for streaming.
+	install, err := c.client.Install(ctx)
+	if err != nil {
+		return err
+	}
+	defer install.CloseSend()
```
samribeiro, line 74: no need to do this, the treams will be finished gracefully by gRPC when this function returns.

gnoi/os/client.go
```diff
@@ -27,11 +30,170 @@ import (
...
+		return fmt.Errorf("No reader passed to client")
+	}
+	file, fileSize, fileClose, err := c.read(imgPath)
+	if err != nil {
+		return err
+	}
+	defer fileClose()
+
+	// Create Install client for streaming.
+	install, err := c.client.Install(ctx)
```
samribeiro, line 70: repackage this context with a cancel such that you can cancel it and avoid blocked Send/Recv:
https://godoc.org/context#WithCancel

gnoi/os/client.go
```diff
@@ -27,11 +30,170 @@ import (
 // Client handles requesting OS RPCs.
 type Client struct {
 	client pb.OSClient
+	read   func(string) (file io.ReaderAt, size uint64, close func() error, err error)
```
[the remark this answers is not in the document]
ericm: I added it to the struct so i can pass in a mock fileReader in the tests during the construction of the client. Is there a better way to pass in the mock?

gnoi/os/client.go
```diff
@@ -27,11 +30,170 @@ import (
...
+	}
+	transferResp, err := install.Recv()
+	if err != nil {
+		return err
+	}
+	_, validated, err := validateInstallResponse(transferResp)
+	if err != nil {
+		return err
+	}
+	if validated {
```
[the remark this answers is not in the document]
ericm: Should we just return nil here or should we notify the user using log?

gnoi/os/client.go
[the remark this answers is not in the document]
samribeiro: yep, something like this inside your test:

defaultFileReader := fileReader
fileReader  = func(path string) (...) {
    ...
}
defer func() { fileReader = defaultFileReader }()

ericm:
Added fileReader change as well

gnoi/os/client.go
```diff
@@ -18,22 +18,177 @@ package os
...
+	go func() {
+		for {
+			response, err := install.Recv()
+			if err != nil {
+				errs <- err
+				return
+			}
+			switch resp := response.Response.(type) {
+			case *pb.InstallResponse_TransferProgress:
+				if printStatus {
```
samribeiro, line 120: I think you can move this printStatus as a flag in utils too

samribeiro: and here just call a function in utils.

samribeiro:
tests? =)

ericm:
Added PrintStatus. I was thinking of making a new PR for the tests after this is merged since they're pretty long.

ericm:
Build error will be fixed in #136

gnoi/os/client.go
samribeiro: needs updating here?

utils/utils.go
```diff
@@ -11,15 +11,23 @@ import (
...
+	printStatus = flag.Bool("print_status", false, "Print status of file transfer.")
 )
 
 // PrintProto prints a Proto in a structured way.
 func PrintProto(m proto.Message) {
 	fmt.Println(FormatProto(m))
 }
 
+// PrintStatus prints the percentage transferred if print_status is true
+func PrintStatus(transferred, fileSize uint64) {
```
samribeiro, line 25: Just give it the string directly

utils/utils.go
```diff
@@ -11,15 +11,23 @@ import (
 )
 
 var (
-	usePretty = flag.Bool("pretty", false, "Shows PROTOs using Pretty package instead of PROTO Text Marshal")
-	logProto  = flag.Bool("log_proto", false, "If true it prints all sent and received PROTO messages")
+	usePretty   = flag.Bool("pretty", false, "Shows PROTOs using Pretty package instead of PROTO Text Marshal")
+	logProto    = flag.Bool("log_proto", false, "If true it prints all sent and received PROTO messages")
+	printStatus = flag.Bool("print_status", false, "Print status of file transfer.")
```
samribeiro, line 16: printProgress

Issue #108, opened 2020-07-02
Refactor certificate usage for the binaries.


samribeiro:
- CA.crt -> Certificate Authority certificate (usually a self-signed certificate that signs all the others).
- CA.key -> CA Private key (usually should be kept very secret).
- client.crt/target.crt -> A client or target certificate that was signed by the CA.
- client.key/target.key -> The private key corresponding to the client or target.

### Before Authentication

- The client has: client.crt, client.key, CA.crt
- The target has: target.crt, target.key, CA.crt

### During Authentication

- The client sends the client certificate to the target.
- The target sends the target certificate to the client.
- The client uses the CA certificate to validate that the received target certificate is valid.
- The target uses the CA certificate to validate that the received client certificate is valid.
- The client verifyes that the name in the target certificate matches with the expected one.

### After Authentication

- The private keys are used for establishing the secure channels.

## Current certificate usage for the binaries

The following binaries receive ca.crt, client.crt/target.crt, client.key/target.key:
gnmi_get
gnmi_set
gnmi_target
gnoi_reset

However, gnoi_cert, receives ca.crt and ca.key. This is usefull because this client needs to sign and install certificates in the target. In addition, gnoi_cert uses this ca.crt and key to dynamically create a client.crt to connect to the target.

## Proposed certificate usage for all binaries

Have all binaries accept one of the two options:
1) ca.crt, client.crt/target.crt, client.key/target.key;
or
2) ca.key, ca.crt;

Package these two options in a library under utils where it can be used by all other binaries.

ericm:
Do we want a flag to represent switching between the 2 options or just use the first one if the cert flag is set

samribeiro:
if ca.crt, client.crt/target.crt, client.key/target.key are provided, use option 1.
if only ca.key and ca.crt are provided, use option 2 (what gnoi cert does).
print usage message otherwise.

PR #133, opened 2020-07-09
Refactored credential utils to accept ca.crt, client.crt, client.key or ca.crt, ca.key

ericm:
Related issue: #108 
This will allow clients that use the library to avail of both options :)

### DIFF
--- a/utils/credentials/credentials.go
+++ b/utils/credentials/credentials.go
@@ -29,6 +29,7 @@ import (
 	"google.golang.org/grpc/metadata"
 
 	log "github.com/golang/glog"
+	"github.com/google/gnxi/utils/entity"
 )
 
 var (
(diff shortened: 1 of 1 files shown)


utils/credentials/credentials.go
```diff
@@ -87,6 +84,33 @@ func LoadCertificates() ([]tls.Certificate, *x509.CertPool) {
...
+		log.Exitf("Failed to create a signed entity: %v", err)
+	}
+	caPool := x509.NewCertPool()
+	caPool.AddCert(caEnt.Certificate.Leaf)
+	return []tls.Certificate{*clientEnt.Certificate}, caPool
+}
+
+// LoadCertificates loads certificates from files or from the CA.
+func LoadCertificates() ([]tls.Certificate, *x509.CertPool) {
+	if *ca != "" && *key != "" {
```
samribeiro, line 104: there needs to be a differentiation between the two keys. There is one for the CA and one for the client/target. This is important to highlight.

	ca             = flag.String("ca", "", "CA certificate file.")
	caKey             = flag.String("ca_key", "", "CA private key file.")
	cert           = flag.String("cert", "", "Certificate file.")
	key            = flag.String("key", "", "Private key file.")

ericm:
Also added entity to return so that it can be used gnoi_cert

PR #134, opened 2020-07-10
Added logging to Verify client RPC

ericm:
Added Verify RPC to client and added logging

### DIFF
--- a/gnoi/os/client.go
+++ b/gnoi/os/client.go
@@ -62,10 +62,13 @@ func (c *Client) Activate(ctx context.Context, version string) error {
 
 // Verify invokes the Verify RPC for the OS service.
 func (c *Client) Verify(ctx context.Context) (version, activationFailMsg string, err error) {
+	request := &pb.VerifyRequest{}
+	utils.LogProto(request)
 	var out *pb.VerifyResponse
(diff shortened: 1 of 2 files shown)


gnoi/os/client.go
```diff
@@ -62,10 +62,13 @@ func (c *Client) Activate(ctx context.Context, version string) error {
 
 // Verify invokes the Verify RPC for the OS service.
 func (c *Client) Verify(ctx context.Context) (version, activationFailMsg string, err error) {
+	request := &pb.VerifyRequest{}
+	utils.LogProto(request)
 	var out *pb.VerifyResponse
-	if out, err = c.client.Verify(ctx, &pb.VerifyRequest{}); err != nil {
+	if out, err = c.client.Verify(ctx, request); err != nil {
```
loriscn12, line 68: pass &pb.VerifyRequest{} directly, no var assignment needed IMO.

ericm: I'm not sure how I'll pass the request into `LogProto` when not assigning it

loriscn12: would the following not work?

c.client.Verify(ctx, &pb.VerifyRequest{})

loriscn12: sorry i missed the logging part :P

PR #135, opened 2020-07-13
gNOI Target OS Install

### LINKED ISSUE #112: Implement OS Install RPC for mock target

Jared-Mullin:
Closes #112


gnoi/os/server.go
```diff
@@ -51,3 +59,84 @@ func (s *Server) Activate(ctx context.Context, request *pb.ActivateRequest) (*pb
...
+	var resp *pb.InstallRequest
+	var err error
+	if resp, err = stream.Recv(); err != nil {
+		return err
+	}
+	transferRequest := resp.GetTransferRequest()
+	if transferRequest == nil {
+		return errors.New("failed to receive TransferRequest")
+	}
+	if version := transferRequest.Version; s.manager.IsInstalled(version) {
```
samribeiro, line 74: also handle the following errors:
 INSTALL_RUN_PACKAGE
INSTALL_IN_PROGRESS

gnoi/os/server.go
```diff
@@ -51,3 +59,84 @@ func (s *Server) Activate(ctx context.Context, request *pb.ActivateRequest) (*pb
 	}
 	return &pb.ActivateResponse{Response: &pb.ActivateResponse_ActivateOk{}}, nil
 }
+
+// Install receives an OS package, validates the package and then installs the package.
+func (s *Server) Install(stream pb.OS_InstallServer) error {
+	var resp *pb.InstallRequest
+	var err error
+	if resp, err = stream.Recv(); err != nil {
```
samribeiro, line 67: Worth logging the sends and receives that are not InstallRequest->bytes with the utils LogProto

Jared-Mullin: Sounds good, will do

gnoi/os/server.go
```diff
@@ -51,3 +61,118 @@ func (s *Server) Activate(ctx context.Context, request *pb.ActivateRequest) (*pb
...
+	return nil
+}
+
+// ReceiveOS receives and parses requests from stream, storing OS package into a buffer.
+func ReceiveOS(stream pb.OS_InstallServer, errorChan chan error, updateProgress chan uint64, transferredOS chan *bytes.Buffer) {
+	bb := new(bytes.Buffer)
+	prev := 0
+	for {
+		in, err := stream.Recv()
+		if err == io.EOF {
```
samribeiro, line 153: this is not used, we use the TransferEnd to singal the end of the content.

gnoi/os/server.go
```diff
@@ -51,3 +61,118 @@ func (s *Server) Activate(ctx context.Context, request *pb.ActivateRequest) (*pb
...
+
+	response = &pb.InstallResponse{Response: &pb.InstallResponse_TransferReady{TransferReady: &pb.TransferReady{}}}
+	utils.LogProto(response)
+	if err = stream.Send(response); err != nil {
+		return err
+	}
+	errorChan := make(chan error)
+	updateProgress := make(chan uint64)
+	transferredOS := make(chan *bytes.Buffer)
+	go ReceiveOS(stream, errorChan, updateProgress, transferredOS)
```
samribeiro, line 115: =) You look like me the first time I played with channels. Let me ask you a simple question and perhaps that will lead you to simplify this code. Why do you need a go thread here? ;) At this point all you need to do is, receive data, receive transfer end or handle a stream error.

samribeiro: - receive TransferRequest
- check version is the one running
- check if another install is already in progress, else quit with error
- lock this as being the only install, all others will fail
- received next response (transfer content or transfer end or error)
- if transfer content totals an amount of chunks then log the amount
- if error, then finish
- if transfer end then validate image and return success
- release the lock

samribeiro: forgot to include the "send the sync progress" wich you can interleave in between every reception.

gnoi/os/server.go
```diff
@@ -13,30 +13,42 @@ limitations under the License.
...
+const (
+	chunkSize = 1000000
+)
+
 // Server is an OS Management service.
 type Server struct {
 	pb.OSServer
-	manager *Manager
+	manager     *Manager
+	InstallLock chan bool
```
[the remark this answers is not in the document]
samribeiro: perhaps installToken fits better

gnoi/os/server.go
```diff
@@ -62,3 +74,107 @@ func (s *Server) Verify(ctx context.Context, _ *pb.VerifyRequest) (*pb.VerifyRes
...
+			return err
+		}
+		return nil
+	}
+	select {
+	case <-s.InstallLock:
+		defer func() {
+			s.InstallLock <- true
+		}()
+		break
```
samribeiro, line 118: no need for break in cases

gnoi/os/server.go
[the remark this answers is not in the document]
Jared-Mullin: That makes sense to me! Thanks

gnoi/os/server.go
Jared-Mullin: Thanks for catching that!

gnoi/os/server.go
```diff
@@ -62,3 +73,106 @@ func (s *Server) Verify(ctx context.Context, _ *pb.VerifyRequest) (*pb.VerifyRes
...
+// ReceiveOS receives and parses requests from stream, storing OS package into a buffer, and updating the progress.
+func ReceiveOS(stream pb.OS_InstallServer) (*bytes.Buffer, error) {
+	bb := new(bytes.Buffer)
+	prev := 0
+	for {
+		in, err := stream.Recv()
+		if err != nil {
+			return nil, err
+		}
+		utils.LogProto(in)
```
samribeiro, line 158: Only log the proto when it is not a data transfer else we flood screen/log =)

gnoi/os/server.go
```diff
@@ -62,3 +73,106 @@ func (s *Server) Verify(ctx context.Context, _ *pb.VerifyRequest) (*pb.VerifyRes
...
+		}
+		return err
+	}
+	s.manager.Install(mockOS.Version, mockOS.ActivationFailMessage)
+	return nil
+}
+
+// ReceiveOS receives and parses requests from stream, storing OS package into a buffer, and updating the progress.
+func ReceiveOS(stream pb.OS_InstallServer) (*bytes.Buffer, error) {
+	bb := new(bytes.Buffer)
```
samribeiro, line 151: bb := bytes.Buffer{}

just for consistency

Jared-Mullin: Agree! Changed!

gnoi/os/server.go
samribeiro: I think you removed the wrong log! =)
The transfer progress is fine. The receiving bytes from the client can go.

Jared-Mullin: Definitely, my bad, sorry!

PR #139, opened 2020-07-16
Refactored gnoi_cert for use with client cert + key

### LINKED ISSUE #108: Refactor certificate usage for the binaries.

ericm:
Related issue: Closes #108

### DIFF
--- a/gnoi_cert/gnoi_cert.go
+++ b/gnoi_cert/gnoi_cert.go
@@ -63,22 +63,21 @@ func main() {
 		log.Exit("Must set a Common Name ID with -target_name.")
 	}
 
-	if *certID == "" {
-		log.Exit("Must set a certificate ID with -cert_id.")
-	}
-
 	ctx, cancel = context.WithTimeout(context.Background(), *timeOut)
 	defer cancel()
 
 	switch *op {
 	case "provision":
 		caEnt = credUtils.GenerateCA()
+		certIDCheck()
 		provision()
 	case "install":
 		caEnt = credUtils.GenerateCA()
+		certIDCheck()
 		install()
 	case "rotate":
 		caEnt = credUtils.GenerateCA()
+		certIDCheck()
 		rotate()
 	case "revoke":
 		revoke()
@@ -88,6 +87,12 @@ func main() {
 		get()
 	default:
 		log.Exitf("Unknown operation: %q", *op)
+	}
+}
+
+func certIDCheck() {
+	if *certID == "" {
+		log.Exit("Must set a certificate ID with -cert_id.")
 	}
 }
 
--- a/gnoi_cert/gnoi_cert.go
+++ b/gnoi_cert/gnoi_cert.go
@@ -68,15 +68,15 @@ func main() {
 
 	switch *op {
 	case "provision":
-		caEnt = credUtils.GenerateCA()
+		caEnt = credUtils.GetCAEntity()
 		certIDCheck()
 		provision()
 	case "install":
-		caEnt = credUtils.GenerateCA()
+		caEnt = credUtils.GetCAEntity()
 		certIDCheck()
 		install()
 	case "rotate":
-		caEnt = credUtils.GenerateCA()
+		caEnt = credUtils.GetCAEntity()
 		certIDCheck()
 		rotate()
 	case "revoke":
--- a/gnoi_cert/gnoi_cert.go
+++ b/gnoi_cert/gnoi_cert.go
@@ -139,12 +139,16 @@ func gnoiAuthenticated(targetName string) (*grpc.ClientConn, *cert.Client) {
 		signedCert = *clientEnt.Certificate
 	}
 
-	caFile, err := ioutil.ReadFile(*ca)
-	if err != nil {
-		log.Exitf("Couldn't read CA file: %v", err)
-	}
 	caPool := x509.NewCertPool()
-	caPool.AppendCertsFromPEM(caFile)
+	if caEnt != nil {
+		caPool.AddCert(caEnt.Certificate.Leaf)
+	} else {
+		caFile, err := ioutil.ReadFile(*ca)
+		if err != nil {
+			log.Exitf("Couldn't read CA file: %v", err)
+		}
+		caPool.AppendCertsFromPEM(caFile)
+	}
 
 	opts := []grpc.DialOption{grpc.WithTransportCredentials(credentials.NewTLS(
 		&tls.Config{
--- a/gnoi_cert/gnoi_cert.go
+++ b/gnoi_cert/gnoi_cert.go
@@ -23,9 +23,10 @@ import (
 	"crypto/x509/pkix"
 	"flag"
 	"fmt"
+	"io/ioutil"
 	"reflect"
-	"time"
 	"strings"
+	"time"
 
 	"github.com/google/gnxi/gnoi/cert"
 	"github.com/google/gnxi/utils/entity"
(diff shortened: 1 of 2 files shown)


utils/credentials/credentials.go
```diff
@@ -125,7 +125,7 @@ func ClientCredentials(server string) []grpc.DialOption {
 		if *insecure {
 			tlsConfig.InsecureSkipVerify = true
 		} else {
-			certificates, certPool, _ := LoadCertificates()
+			certificates, certPool := LoadCertificates()
```
loriscn12, line 128: wasn't this fixed at: https://github.com/google/gnxi/pull/136/files ?

ericm: Yeah I think I branched from the PR so it was included in the diff here

ericm:
> Thanks Eric, LGTM from a syntax point of view, are you planning to add tests?

I can if you think it's necessary. The binaries don't usually seem have tests but in this case it might need some.

gnoi_cert/gnoi_cert.go
```diff
@@ -122,17 +127,36 @@ func gnoiEncrypted(c tls.Certificate) (*grpc.ClientConn, *cert.Client) {
 
 // gnoiAuthenticated creates an authenticated TLS connection to the target.
 func gnoiAuthenticated(targetName string) (*grpc.ClientConn, *cert.Client) {
-	clientEnt, err := entity.CreateSigned("client", nil, caEnt)
-	if err != nil {
-		log.Exitf("Failed to create a signed entity: %v", err)
+	var signedCert tls.Certificate
```
samribeiro, line 130: This logic already existds here? I think you can use it unless I am missing something:
https://github.com/google/gnxi/blob/master/utils/credentials/credentials.go#L103

gnoi_cert/gnoi_cert.go
```diff
@@ -60,30 +63,25 @@ var (
...
 
-	if *ca == "" || *key == "" {
-		log.Exit("-ca and -key must be set with file locations")
-	}
 	if *targetCN == "" {
 		log.Exit("Must set a Common Name ID with -target_name.")
 	}
-
-	var err error
-	if caEnt, err = entity.FromFile(*ca, *key); err != nil {
```
samribeiro, line 71: You can add a function to cretdentials to return an entity if already created here:
https://github.com/google/gnxi/blob/master/utils/credentials/credentials.go#L89

this function would fail if CA or CAKey are not present

gnoi_cert/gnoi_cert.go
```diff
@@ -41,7 +42,9 @@ var (
 	certIDs    = flag.String("cert_ids", "", "Comma separated list of Certificate Management certificate IDs for revoke operation")
 	op         = flag.String("op", "get", "Certificate Management operation, one of: provision, install, rotate, get, revoke, check")
 	ca         = flag.String("ca", "", "CA certificate file.")
-	key        = flag.String("key", "", "Private key file.")
+	caKey      = flag.String("ca_key", "", "CA private key file.")
```
samribeiro, line 45: given the suggested below you wont even need to have any key or certificates requested here. All can be done in credentials.go

gnoi_cert/gnoi_cert.go
ericm: I could but use it but I'd have to add some return paths that won't be used by other clients (for caEnt)

samribeiro: yep, as suggested above! =)

gnoi_cert/gnoi_cert.go
```diff
@@ -60,30 +59,26 @@ var (
...
+	if *certID == "" {
+		log.Exit("Must set a certificate ID with -cert_id.")
 	}
 
 	ctx, cancel = context.WithTimeout(context.Background(), *timeOut)
 	defer cancel()
 
 	switch *op {
 	case "provision":
-		certIDCheck()
```
samribeiro, line 80: certIDChecks are still needed.

gnoi_cert/gnoi_cert.go
```diff
@@ -60,30 +59,26 @@ var (
...
-		log.Exit("-ca and -key must be set with file locations")
-	}
 	if *targetCN == "" {
 		log.Exit("Must set a Common Name ID with -target_name.")
 	}
 
-	var err error
-	if caEnt, err = entity.FromFile(*ca, *key); err != nil {
-		log.Exitf("Failed to load certificate and key from file: %v", err)
+	if *certID == "" {
```
samribeiro, line 66: if revoke operation this is not needed

samribeiro: as the certIDs can be provided isntead

utils/credentials/credentials.go
```diff
@@ -139,6 +143,18 @@ func ClientCredentials(server string) []grpc.DialOption {
 	return opts
 }
 
+// GenerateCA generates a CA entity from a CA file and private key.
+func GenerateCA() *entity.Entity {
```
samribeiro, line 147: GetCAEntity

utils/credentials/credentials.go
```diff
@@ -139,6 +143,18 @@ func ClientCredentials(server string) []grpc.DialOption {
...
+	if *caKey == "" {
+		log.Exit("-ca_key must be set with file locations")
+	}
+	var err error
+	if caEnt, err = entity.FromFile(*ca, *caKey); err != nil {
```
samribeiro, line 152: if caEnt already exists, just return it

utils/credentials/credentials.go
```diff
@@ -86,9 +87,12 @@ func loadCerts() ([]tls.Certificate, *x509.CertPool) {
 
 // generateFromCA generates a client certificate from the provided CA.
 func generateFromCA() ([]tls.Certificate, *x509.CertPool) {
-	caEnt, err := entity.FromFile(*ca, *caKey)
-	if err != nil {
-		log.Exitf("Failed to load certificate and key from file: %v", err)
+	if caEnt == nil {
```
samribeiro, line 90: just call GetCAEntity()

PR #141, opened 2020-07-17
Tests for Install Service

Jared-Mullin:
Related: #147 
What's Tested:
Valid new OS that's compatible
Initial Request that isn't the TransferRequest
Force transferring currently running OS
Transferring an already installed OS
Transferring an OS with a bad has (Integrity Check Fails)
Transferring an Incompatible OS
Transferring a random collection of bytes (Parse Fail)
Attempting to install while another goroutine is installing

### DIFF
--- a/gnoi/os/server_test.go
+++ b/gnoi/os/server_test.go
@@ -65,6 +65,7 @@ func (m mockTransferStream) Recv() (*pb.InstallRequest, error) {
 				out, _ = proto.Marshal(&m.os.MockOS)
 			} else {
 				out = make([]byte, 10000000)
+				rand.Read(out)
 			}
 			return &pb.InstallRequest{Request: &pb.InstallRequest_TransferContent{TransferContent: out}}, nil
 		}
--- a/gnoi/os/server_test.go
+++ b/gnoi/os/server_test.go
@@ -188,10 +188,12 @@ func TestTargetReceiveOS(t *testing.T) {
 	}}
 	oS.Hash()
 	tests := []struct {
+		name   string
 		stream *mockTransferStream
 		err    error
 	}{
 		{
+			name: "sending TransferContent request",
 			stream: &mockTransferStream{
 				response: make(chan *pb.InstallResponse, 1),
 				os:       oS,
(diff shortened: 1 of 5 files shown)


gnoi/os/server_test.go
```diff
@@ -151,28 +179,196 @@ func TestTargetActivateAndVerify(t *testing.T) {
...
+		}
+		close(test.stream.result)
+		got.res = <-test.stream.result
+		if diff := pretty.Compare(test.want, got); diff != "" {
+			t.Errorf("Install(stream pb.OS_InstallServer): (-want +got):\n%s", diff)
+		}
+	}
+}
+
+func TestMultipleInstalls(t *testing.T) {
```
loriscn12, line 344: because this test does not have multiple named cases, I would add a comment to clarify what tests and why it is important.

gnoi/os/server_test.go
```diff
@@ -151,28 +179,196 @@ func TestTargetActivateAndVerify(t *testing.T) {
...
+					Version: "1.0.2c",
+				}},
+			},
+			want: &installResult{
+				res: &pb.InstallResponse{Response: &pb.InstallResponse_InstallError{InstallError: &pb.InstallError{Type: pb.InstallError_PARSE_FAIL}}},
+				err: nil,
+			},
+		},
+	}
+	for _, test := range tests {
```
loriscn12, line 331: use https://github.com/google/gnxi/blob/4abd369b790998675df042f5945979d9c9d1e027/gnoi/os/client_test.go#L330

context: https://golang.org/pkg/testing/

gnoi/os/server_test.go
```diff
@@ -151,28 +179,196 @@ func TestTargetActivateAndVerify(t *testing.T) {
...
+	}}
+	oS.Hash()
+	incompatibleOS := &mockos.OS{MockOS: mockosPb.MockOS{
+		Version:      "1.0.2b",
+		Cookie:       "cookiestring",
+		Padding:      buf,
+		Incompatible: true,
+	}}
+	incompatibleOS.Hash()
+	tests := []struct {
```
loriscn12, line 235: make sure you have a string field called name or similar(this has to do with my next comment)about t.Run)

gnoi/os/server_test.go
Jared-Mullin: Sure, that makes sense!

gnoi/os/server_test.go
Jared-Mullin: Totally! Adding now

gnoi/os/server_test.go
Jared-Mullin: Added

PR #154, opened 2020-07-20
Update individual README files

samribeiro:


gnoi_os/README.md
```diff
@@ -4,13 +4,11 @@ A simple shell binary that performs OS client operations against a gNOI target.
...
-* `-op install` installs the provided OS image onto the target when it doesn't already 
-    have that OS version installed
+* `-op install` installs the provided OS image onto the target.
 
-* `-op activate` tells the target to boot into the specified OS version on next reboot if
-    it's installed.
+* `-op activate` tells the target to boot into the specified OS version on next reboot.
 
-* `-op verify` verifies the version of the OS currenly running on the target
+* `-op verify` verifies the version of the OS currenly running on the target.
```
Jared-Mullin, line 11: nit: currently

README.md
```diff
@@ -31,33 +50,47 @@ _Note_: These tools are intended for testing and as reference implementation of
...
+./../src/github.com/google/gnxi/certs/generate.sh
 ```
 
-## Docker
+### Running a client
 
-[FAUCET](https://github.com/faucetsdn/gnmi) currently includes a [Dockerfile](https://github.com/faucetsdn/gnmi/blob/master/Dockerfile) to setup the environment that facilitates testing these tools against network equipment.
+```
+cd $GOPATH\bin
+./gnoi_reset \
```
Jared-Mullin, line 85: Should we remove the "./" since the other READMEs don't have it?

README.md
```diff
@@ -31,33 +50,47 @@ _Note_: These tools are intended for testing and as reference implementation of
...
 
-### Running
+### Building and installing binaries
+
+```
+cd $GOPATH
+mkdir bin
+go install github.com/google/gnxi/...
+ls -la $GOPATH\bin
+```
```
Jared-Mullin, line 72: Should we recommend adding $GOPATH/bin to their path? So that they can run it anywhere, like:
gnoi_os

README.md
```diff
@@ -31,33 +50,47 @@ _Note_: These tools are intended for testing and as reference implementation of
...
 ```
 
-### Running
+### Building and installing binaries
+
+```
+cd $GOPATH
+mkdir bin
+go install github.com/google/gnxi/...
+ls -la $GOPATH\bin
```
Jared-Mullin, line 71: Should we recommend the user add $GOPATH/bin to their $PATH? So that they can use the binaries anywhere

README.md
Jared-Mullin, line 85: Should we add "./" to the other READMEs? It seems to be absent from a few

README.md
samribeiro: yes, thanks

README.md
samribeiro: yes, thanks

PR #155, opened 2020-07-20
Added certificate refactor to gnoi_target

### LINKED ISSUE #153: gnoi_target if certificates are provided (ca, target & target_key) start in provisioned mode
Right now, gnoi_target is starting in bootstrapping mode by default.
However it should be possible to provide target certificates such that it starts in provisioned instead.

ericm:
Closes #153

### DIFF
--- a/gnoi/cert/manager.go
+++ b/gnoi/cert/manager.go
@@ -62,12 +62,16 @@ type Manager struct {
 	mu        sync.RWMutex
 }
 
+var generatePrivateKey = func() (*rsa.PrivateKey, error) {
+	return rsa.GenerateKey(rand.Reader, RSABitSize)
+}
+
 // NewManager returns a Manager.
 func NewManager(settings *Settings) *Manager {
 	var (
 		privateKey crypto.PrivateKey
-		caBundle   []*x509.Certificate
-		certInfo   map[string]*Info
+		certInfo   =  map[string]*Info{}
+		caBundle    = []*x509.Certificate{}
 	)
 	if settings.Cert != nil {
 		privateKey = settings.Cert.PrivateKey
@@ -75,7 +79,7 @@ func NewManager(settings *Settings) *Manager {
 		caBundle = []*x509.Certificate{settings.CA}
 	} else {
 		var err error
-		privateKey, err = rsa.GenerateKey(rand.Reader, RSABitSize)
+		privateKey, err = generatePrivateKey()
 		if err != nil {
 			log.Exit("couldn't generate private key for manager: ", err)
 		}
--- a/gnoi/cert/manager.go
+++ b/gnoi/cert/manager.go
@@ -52,11 +52,11 @@ type Manager struct {
 }
 
 // NewManager returns a Manager.
-func NewManager(privateKey crypto.PrivateKey) *Manager {
+func NewManager(privateKey crypto.PrivateKey, caBundle []*x509.Certificate) *Manager {
 	return &Manager{
 		privateKey: privateKey,
 		certInfo:   map[string]*Info{},
-		caBundle:   []*x509.Certificate{},
+		caBundle:   caBundle,
 		locks:      map[string]bool{},
 		notifiers:  []Notifier{},
 	}
--- a/gnoi/cert/manager.go
+++ b/gnoi/cert/manager.go
@@ -40,6 +40,13 @@ type Info struct {
 // Notifier is called with number of Certificates and CA Certificates.
 type Notifier func(int, int)
 
+// ManagerSettings contains the certs and CA pool to be passed into the Manager.
+type ManagerSettings struct {
+	CertID   string
+	Cert     *tls.Certificate
+	CABundle []*x509.Certificate
+}
+
 // Manager manages Certificates and CA Bundles.
 type Manager struct {
 	privateKey crypto.PrivateKey
@@ -52,11 +59,20 @@ type Manager struct {
 }
 
 // NewManager returns a Manager.
-func NewManager(privateKey crypto.PrivateKey, caBundle []*x509.Certificate) *Manager {
+func NewManager(settings *ManagerSettings) *Manager {
+	if settings == nil {
+		return &Manager{}
+	}
+	certInfo := map[string]*Info{}
+	var privateKey crypto.PrivateKey
+	if settings.Cert != nil {
+		privateKey = settings.Cert.PrivateKey
+		certInfo = map[string]*Info{settings.CertID: {certID: settings.CertID, cert: settings.Cert.Leaf}}
+	}
 	return &Manager{
 		privateKey: privateKey,
-		certInfo:   map[string]*Info{},
-		caBundle:   caBundle,
+		certInfo:   certInfo,
+		caBundle:   settings.CABundle,
 		locks:      map[string]bool{},
 		notifiers:  []Notifier{},
 	}
--- a/gnoi/cert/manager.go
+++ b/gnoi/cert/manager.go
@@ -40,8 +40,8 @@ type Info struct {
 // Notifier is called with number of Certificates and CA Certificates.
 type Notifier func(int, int)
 
-// ManagerSettings contains the certs and CA pool to be passed into the Manager.
-type ManagerSettings struct {
+// Settings contains the certs and CA pool to be passed into the Manager.
+type Settings struct {
 	CertID   string
 	Cert     *tls.Certificate
 	CABundle []*x509.Certificate
@@ -59,7 +59,7 @@ type Manager struct {
 }
 
 // NewManager returns a Manager.
-func NewManager(settings *ManagerSettings) *Manager {
+func NewManager(settings *Settings) *Manager {
 	if settings == nil {
 		return &Manager{}
 	}
--- a/gnoi/cert/manager.go
+++ b/gnoi/cert/manager.go
@@ -18,6 +18,7 @@ package cert
 import (
 	"crypto"
 	"crypto/rand"
+	"crypto/rsa"
 	"crypto/tls"
 	"crypto/x509"
 	"crypto/x509/pkix"
@@ -29,6 +30,9 @@ import (
 
 	log "github.com/golang/glog"
 )
+
+// RSABitSize is the size of the required RSA Private Key.
+const RSABitSize = 2048
 
 // Info contains information about a x509 Certificate.
 type Info struct {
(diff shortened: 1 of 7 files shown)


samribeiro:
1) Create a function like credentials.LoadCertificates() that returns nil, nil if no certificates were ever given instead of giving error.
2) Create a Settings struct for cert.Manager where you can pass the certificates and the bundle.
3) Note that certificates have the private key: https://godoc.org/crypto/tls#Certificate
4) In gnoi_target.start() before instantiating gNOIServer attempt loading the certificates with 1). Pass those certificates via Settings to gNOIServer.
5) at the end of gnoi_target.start() if the certificates were loaded call notifyCerts(1,1), else notifyCerts(0,0)

gnoi/cert/server.go
```diff
@@ -336,3 +336,7 @@ func (s *Server) CanGenerateCSR(ctx context.Context, request *pb.CanGenerateCSRR
 	}
 	return ret, nil
 }
+
+func (s *Server) AddCertificates(certID string) (certs, caCerts int) {
```
samribeiro, line 340: Hum.. what does this do? =)

gnoi/gnoi.go
```diff
@@ -45,30 +44,28 @@ type Server struct {
...
+		if err != nil {
+			return nil, fmt.Errorf("failed to generate private key: %v", err)
 		}
 		e, err := entity.CreateSelfSigned("gNOI server", privateKey)
 		if err != nil {
 			return nil, fmt.Errorf("failed to create self signed certificate: %v", err)
 		}
-		defaultCertificate = e.Certificate
+		certSettings.Cert = e.Certificate
+		certSettings.CertID = "0"
```
samribeiro, line 58: use the one in certSettings

gnoi_target/gnoi_target.go
```diff
@@ -37,6 +40,7 @@ var (
 	muServe       sync.Mutex
 	bootstrapping bool
 
+	certID               = flag.String("cert_id", "", "Certificate ID for preloaded certificates")
```
samribeiro, line 43: use a default of "default"

gnoi_target/gnoi_target.go
```diff
@@ -99,19 +105,30 @@ func start() {
...
-	bootstrapping = false
-	notifyCerts(0, 0) // Triggers bootstraping mode.
+	bootstrapping = certs != 0 && caCerts != 0
+	notifyCerts(certs, caCerts) // Triggers bootstraping mode.
 }
 
 // notifyReset is called when the factory reset service requires the server
 // to be restarted.
 func notifyReset() {
+	*certID = ""
```
samribeiro, line 131: no need, will use default above

utils/credentials/credentials.go
```diff
@@ -97,8 +105,8 @@ func generateFromCA() ([]tls.Certificate, *x509.CertPool) {
 	return []tls.Certificate{*clientEnt.Certificate}, caPool
 }
 
-// LoadCertificates loads certificates from files or generates them from the CA.
-func LoadCertificates() ([]tls.Certificate, *x509.CertPool) {
+// ParseCertificates gets certificates from files or generates them from the CA.
+func ParseCertificates() ([]tls.Certificate, *x509.CertPool) {
```
samribeiro, line 109: have this return:
([]*x509.Certificate, []*x509.Certificate)

utils/credentials/credentials.go
```diff
@@ -107,10 +115,18 @@ func LoadCertificates() ([]tls.Certificate, *x509.CertPool) {
 			return generateFromCA()
 		}
 	}
-	log.Exit("Please provide -ca & -key or -ca, -cert & -ca_key")
 	return []tls.Certificate{}, &x509.CertPool{}
 }
 
+// LoadCertificates loads certificates from files and exits if there's an error.
+func LoadCertificates() ([]tls.Certificate, *x509.CertPool) {
+	certs, certPool := ParseCertificates()
```
samribeiro, line 123: create the cert pool here

utils/credentials/credentials.go
```diff
@@ -174,6 +193,11 @@ func ServerCredentials() []grpc.ServerOption {
 	}))}
 }
 
+// GetCACert returns the parsed CA cert.
```
samribeiro, line 196: if you return the list of CA certificates above, then you dont need this

utils/credentials/credentials.go
```diff
@@ -123,6 +139,8 @@ func ClientCredentials(server string) []grpc.DialOption {
 			tlsConfig.InsecureSkipVerify = true
 		} else {
 			certificates, certPool := LoadCertificates()
+			if len(certificates) == 0 {
```
samribeiro, line 142: remove

utils/credentials/credentials.go
```diff
@@ -148,6 +166,7 @@ func GetCAEntity() *entity.Entity {
 	if caEnt, err = entity.FromFile(*ca, *caKey); err != nil {
 		log.Exitf("Failed to load certificate and key from file: %v", err)
 	}
+	caCert = caEnt.Certificate.Leaf
```
samribeiro, line 169: no need after the above change

gnoi/cert/manager.go
```diff
@@ -40,6 +40,13 @@ type Info struct {
 // Notifier is called with number of Certificates and CA Certificates.
 type Notifier func(int, int)
 
+// ManagerSettings contains the certs and CA pool to be passed into the Manager.
+type ManagerSettings struct {
+	CertID   string
+	Cert     *tls.Certificate
```
samribeiro, line 46: lets also use *x509.Certificate
see other comments

ericm:
Everything's working now :)

gnoi/cert/manager.go
```diff
@@ -52,11 +59,20 @@ type Manager struct {
 }
 
 // NewManager returns a Manager.
-func NewManager(privateKey crypto.PrivateKey) *Manager {
+func NewManager(settings *Settings) *Manager {
+	if settings == nil {
```
samribeiro, line 63: No need to have this check, we should fail hard if this happens. If we want to have an empty Manager{} we can always create it without calling this function.

ericm:
Removed :)

gnoi/gnoi.go
```diff
@@ -45,30 +44,27 @@ type Server struct {
 }
 
 // NewServer returns a new server that can be used by the mock target.
-func NewServer(privateKey crypto.PrivateKey, defaultCertificate *tls.Certificate, resetSettings *reset.Settings, notifyReset reset.Notifier, osSettings *os.Settings) (*Server, error) {
-	if defaultCertificate == nil {
```
samribeiro, line 49: keep this logic within NewServer. Meaning, always create a private key and use it to generate a default certificate, then throw the private key away. This default certificate is only used to establish an encrypted session.
Pass the Settings untouched and directly to cert.NewManager.

utils/credentials/credentials.go
```diff
@@ -67,38 +68,39 @@ func (a *userCredentials) RequireTransportSecurity() bool {
 }
 
 // loadFromFile loads the certificates from files.
```
samribeiro, line 70: loads a certificate key pair into a tls certificate and a CA certificate into a x509 certificate.

utils/credentials/credentials.go
```diff
@@ -67,38 +68,39 @@ func (a *userCredentials) RequireTransportSecurity() bool {
...
-func loadFromFile() ([]tls.Certificate, *x509.CertPool) {
+func loadFromFile() ([]tls.Certificate, []*x509.Certificate) {
```
samribeiro, line 71: I think we can drop the slices, and jsut have pointers to tls and x509

utils/credentials/credentials.go
```diff
@@ -67,38 +68,39 @@ func (a *userCredentials) RequireTransportSecurity() bool {
...
 	certificate, err := tls.LoadX509KeyPair(*cert, *key)
 	if err != nil {
 		log.Exitf("could not load client key pair: %s", err)
```
samribeiro, line 74:
```
log.Exit("Could not load key/certificate pair from files:", err)
```

utils/credentials/credentials.go
```diff
@@ -67,38 +68,39 @@ func (a *userCredentials) RequireTransportSecurity() bool {
...
 	}
-
-	certPool := x509.NewCertPool()
+	certificate.Leaf, err = x509.ParseCertificate(certificate.Certificate[0])
+	if err != nil {
+		log.Exitf("could not get leaf for certificate: %s", err)
```
samribeiro, line 78:
```
log.Exit("Could not parse x509 certificate from tls certificate:", err)
```

utils/credentials/credentials.go
```diff
@@ -67,38 +68,39 @@ func (a *userCredentials) RequireTransportSecurity() bool {
...
+	if err != nil {
+		log.Exit("Error parsing CA certificate", err)
 	}
-	return []tls.Certificate{certificate}, certPool
+	return []tls.Certificate{certificate}, []*x509.Certificate{caCert}
 }
 
 // generateFromCA generates a client certificate from the provided CA.
-func generateFromCA() ([]tls.Certificate, *x509.CertPool) {
+func generateFromCA() ([]tls.Certificate, []*x509.Certificate) {
```
samribeiro, line 93: same, I think we can drop the slices

utils/credentials/credentials.go
```diff
@@ -107,8 +109,18 @@ func LoadCertificates() ([]tls.Certificate, *x509.CertPool) {
 			return generateFromCA()
 		}
 	}
-	log.Exit("Please provide -ca & -key or -ca, -cert & -ca_key")
-	return []tls.Certificate{}, &x509.CertPool{}
+	return []tls.Certificate{}, []*x509.Certificate{}
```
samribeiro, line 112: can also drop the slices

gnoi/cert/manager.go
```diff
@@ -52,11 +59,17 @@ type Manager struct {
 }
 
 // NewManager returns a Manager.
-func NewManager(privateKey crypto.PrivateKey) *Manager {
+func NewManager(settings *Settings) *Manager {
+	certInfo := map[string]*Info{}
+	var privateKey crypto.PrivateKey
```
samribeiro, line 64: Private key is always required. If tls.certificate is empty, then generate a key.

gnoi/gnoi.go
ericm: I kept the logic the same from before. I'm still generating the private key and passing the resulting certificate to the settings.

utils/credentials/credentials.go
```diff
@@ -43,6 +44,7 @@ var (
 	usernameKey    = "username"
 	passwordKey    = "password"
 	caEnt          *entity.Entity
+	targetName     = "client.com"
```
samribeiro, line 47: we should make another PR turning this into an exportable flag that can be used by all the binaries

ericm: Sure

ericm:
Should be working now :)

PR #175, opened 2020-07-27
Device Connection History

### LINKED ISSUE #168: Allow Target History
Store the details of the targets that the client has previously connected to for easier access in the future

Jared-Mullin:
Closes #168
Acts as follows:
- If target name and address specified, add to history
- If target name without address:
    - Load address already used
    - Set address to :9339
- If target without address:
    - Use last device
    - If no devices in history, error

### DIFF
--- a/gnxi_tester/cmd/targets.go
+++ b/gnxi_tester/cmd/targets.go
@@ -16,31 +16,26 @@ limitations under the License.
 package cmd
 
 import (
+	"fmt"
+
 	"github.com/google/gnxi/gnxi_tester/config"
 	"github.com/spf13/cobra"
 )
 
 var (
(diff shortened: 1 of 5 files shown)


gnxi_tester/cmd/targets.go
```diff
@@ -0,0 +1,41 @@
...
+
+	"github.com/google/gnxi/gnxi_tester/config"
+	"github.com/spf13/cobra"
+)
+
+var (
+	targetsCmd = &cobra.Command{
+		Use:     "targets",
+		Short:   "Displays target connection history.",
+		Example: "gnxi_tester run [test_names]",
```
ericm, line 29: `gnxi_tester targets`

PR #180, opened 2020-07-30
Added Cert Path to Known Targets

Jared-Mullin:
Removed setting default address
When initially using a target the user must provide the target's name, address, a CA and CA Key
Once added to connection history the user may reuse these details either by specifying the target's name, or entering nothing, then the client will use the last target specified
These details are overwritten if the user specifies the target name and any of the details.

### DIFF
--- a/gnxi_tester/config/history.go
+++ b/gnxi_tester/config/history.go
@@ -28,10 +28,9 @@ type Device struct {
 	CaKey   string `mapstructure:"cakey"`
 }
 
-// SetTarget adds any new target to the target history.
+// SetTarget adds any new target to the list of known targets.
 func SetTarget(targetName, targetAddress, ca, caKey string) error {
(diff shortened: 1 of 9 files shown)


gnxi_tester/config/history.go
```diff
@@ -6,28 +21,59 @@ import (
...
 		}
 		return errors.New("No targets in history and no target specified")
 	}
-	if devices[targetName] == "" {
-		if targetAddress == "" {
-			devices[targetName] = defaultPort
-		} else {
-			devices[targetName] = targetAddress
+	if _, exists := devices[targetName]; !exists {
+		if targetAddress == "" || ca == "" || caKey == "" {
```
ericm, line 55: Maybe we should consider storing paths as absolute only. You could use `path.IsAbs` to ensure its absolute and `path.Join` if it's not

Jared-Mullin: Sounds good to me!

PR #185, opened 2020-08-04
Added dockerfiles to config

ericm:
This will dynamically generate the dockerfiles in the .gnxi directory in the users home. They are needed since the docker sdk needs a path to a dockerfile to build a new image on the host system


ericm:
I could add one for the first method if you want :)

PR #190, opened 2020-08-06
Added --files fixes and reduced chunk size in OS Install

ericm:
- Customisable chunk size for OS transfer in OS service (as previous default was too large and cause a transport error)

- Fixed `--files` in the tester and added a prompt if the user didnt provide the file in `--files`

### DIFF
--- a/gnoi/os/client.go
+++ b/gnoi/os/client.go
@@ -50,7 +50,7 @@ type Client struct {
 	client pb.OSClient
 }
 
-const chunkSize = 5000000
+const chunkSize = 4000000 // 4MB
 
 // NewClient returns a new OS service client.
 func NewClient(c *grpc.ClientConn) *Client {
--- a/gnoi/os/client.go
+++ b/gnoi/os/client.go
@@ -50,7 +50,7 @@ type Client struct {
 	client pb.OSClient
 }
 
-const chunkSize = 4000000 // 4MB
+var readChunkSize uint64 = 4000000
 
 // NewClient returns a new OS service client.
 func NewClient(c *grpc.ClientConn) *Client {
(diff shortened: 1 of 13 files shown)


gnoi/os/client.go
```diff
@@ -50,7 +50,7 @@ type Client struct {
 	client pb.OSClient
 }
 
-const chunkSize = 5000000
+const chunkSize = 4000000 // 4MB
```
samribeiro, line 53: I think this can be a flag with a default.

gnoi/os/server.go
```diff
@@ -24,7 +24,7 @@ import (
 )
 
 const (
-	targetChunkSize = 10000000 // 10MB
+	targetChunkSize = 12000000 // 12MB
```
samribeiro, line 27: same, also a flag with a default.

gnoi/os/client.go
samribeiro: read_chunk_size is probably a better name.

gnoi/os/server.go
samribeiro: receiveChunkSizeAck

gnxi_tester/README.md
```diff
@@ -2,15 +2,35 @@
...
 
 ```
 go get github.com/google/gnxi/gnxi_tester
 go install github.com/google/gnxi/gnxi_tester
 ```
 
-## Run
+### Running
+
+Tests are provided in `~/.gnxi.yml`. By default, the following processes & clients will be run:
```
samribeiro, line 16: any pointer for the format of .gnxi.yml ?
will it be automatically generated? Or is there an example?

ericm: It will be auto generated by viper. So the defaults in config/defaults.go will be written to it.

gnoi/os/client_test.go
```diff
@@ -30,6 +30,8 @@ import (
 	"google.golang.org/grpc"
 )
 
+const readChunkSize = 4000000
```
brianokelly1, line 33: nit: Add a comment, for the units and background as to why this is a reasonable value to chose.

Issue #192, opened 2020-08-06
Create gNMI Subscribe Client


Jared-Mullin:
This issue consists of building a client for testing the gNMI Subscribe service implementation.

PR #193, opened 2020-08-06
gNMI Subscribe Client Setup

Jared-Mullin:
Re: Issue #192 
This PR sets up flags, path parsing and constructs the initial SubscribeRequest

### DIFF
--- a/gnmi_subscribe/gnmi_subscribe.go
+++ b/gnmi_subscribe/gnmi_subscribe.go
@@ -50,7 +50,7 @@ var (
 	pbPathFlags       arrayFlags
 	targetAddr        = flag.String("target_addr", ":9339", "The target address in the format of host:port")
 	targetName        = flag.String("target_name", "", "The target name used to verify the hostname returned by TLS handshake")
-	connectionTimeout = flag.Int("timeout", defaultRequestTimeout, "The timeout for a request in seconds, 10 seconds by default, e.g 10s")
+	connectionTimeout = flag.Duration("timeout", defaultRequestTimeout, "The timeout for a request in seconds, 10 seconds by default, e.g 10s")
 	subscriptionOnce  = flag.Bool("once", false, "If true, the target sends values once off")
 	subscriptionPoll  = flag.Bool("poll", false, "If true, the target sends values on request")
 	streamOnChange    = flag.Bool("stream_on_change", false, "If true, the target sends updates on change")
--- a/gnmi_subscribe/gnmi_subscribe.go
+++ b/gnmi_subscribe/gnmi_subscribe.go
@@ -1,0 +1,172 @@
+/* Copyright 2020 Google Inc.
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+    https://www.apache.org/licenses/LICENSE-2.0
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package main
+
+import (
+	"context"
+	"flag"
+	"strings"
+	"time"
+
+	log "github.com/golang/glog"
+	"github.com/golang/protobuf/proto"
+	"github.com/google/gnxi/utils"
+	"github.com/google/gnxi/utils/credentials"
+	"github.com/google/gnxi/utils/xpath"
+	pb "github.com/openconfig/gnmi/proto/gnmi"
+	"google.golang.org/grpc"
+)
+
+type arrayFlags []string
+
+func (i *arrayFlags) String() string {
+	return "my string representation"
+}
+
+func (i *arrayFlags) Set(value string) error {
+	*i = append(*i, value)
+	return nil
+}
+
+const (
+	defaultTimeout = 10
+)
+
+var (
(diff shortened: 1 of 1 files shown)


gnmi_subscribe/gnmi_subscribe.go
```diff
@@ -0,0 +1,172 @@
...
+	"github.com/google/gnxi/utils/credentials"
+	"github.com/google/gnxi/utils/xpath"
+	pb "github.com/openconfig/gnmi/proto/gnmi"
+	"google.golang.org/grpc"
+)
+
+type arrayFlags []string
+
+func (i *arrayFlags) String() string {
+	return "my string representation"
```
ericm, line 33: Is this correct? If it just requires you to implement an interface maybe just return `""`.

Jared-Mullin: Agreed

gnmi_subscribe/gnmi_subscribe.go
```diff
@@ -0,0 +1,172 @@
...
+)
+
+var (
+	subscribeClient    pb.GNMI_SubscribeClient
+	xPathFlags         arrayFlags
+	pbPathFlags        arrayFlags
+	targetAddr         = flag.String("target_addr", ":9339", "The target address in the format of host:port")
+	targetName         = flag.String("target_name", "", "The target name used to verify the hostname returned by TLS handshake")
+	connectionTimeout  = flag.Duration("timeout", defaultTimeout, "The timeout for a request, 10 seconds by default")
+	valueStreamingMode = flag.String("value_mode", "STREAM", "The mode of streaming values the target should use use, STREAM by default")
```
samribeiro, line 52: I would define instead two booleans.
-once
-pool
 If none is mentioned then use stream (default). If both are defined, send error that only one of either can be used.

Jared-Mullin: That's a lot cleaner, thank you

gnmi_subscribe/gnmi_subscribe.go
```diff
@@ -0,0 +1,172 @@
...
+	subscriptionMode   = flag.String("subscription_mode", "TARGET_DEFINED", "The subscription mode the target should use, TARGET_DEFINED by default")
```
samribeiro, line 53: same, if -stream_on_change is defined then use ON_CHANGE. If not defined, use TARGET_DEFINED.
If -sample_interval is defined use SAMPLE.

gnmi_subscribe/gnmi_subscribe.go
```diff
@@ -0,0 +1,172 @@
...
+	encodingFormat     = flag.String("encoding", "JSON_IETF", "The encoding format used by the target for notifications")
+	allowAggregation   = flag.Bool("allow_aggregation", false, "If true, the elements which are marked eligible are aggregated")
```
samribeiro, line 55: no need to support this.

gnmi_subscribe/gnmi_subscribe.go
samribeiro: also add:
 bool suppress_redundant (default false)
 uint64 heartbeat_interval ((default 10x sample_interval)
This is used if SAMPLE mode is used.

gnmi_subscribe/gnmi_subscribe.go
```diff
@@ -0,0 +1,192 @@
...
+	var heartbeat uint64
+	switch {
+	case *streamOnChange && *sampleInterval != 0:
+		return nil, errors.New("Only one of -stream_on_change and -sample_interval can be set")
+	case *streamOnChange:
+		subscriptionMode = pb.SubscriptionMode_ON_CHANGE
+	case *sampleInterval != 0:
+		subscriptionMode = pb.SubscriptionMode_SAMPLE
+		if *heartbeatInterval == 0 {
+			heartbeat = *sampleInterval * 10
```
samribeiro, line 174: After thinking about it, I take it back, lets not enforce a default. Use it only if set.

Jared-Mullin: Sounds good!

gnmi_subscribe/gnmi_subscribe.go
```diff
@@ -0,0 +1,186 @@
...
+	"github.com/google/gnxi/utils/credentials"
+	"github.com/google/gnxi/utils/xpath"
+	"github.com/openconfig/gnmi/proto/gnmi"
+	pb "github.com/openconfig/gnmi/proto/gnmi"
+	"google.golang.org/grpc"
+)
+
+type arrayFlags []string
+
+func (i *arrayFlags) String() string {
```
brianokelly1, line 34: I might be a touch unfamiliar with the code base (I see this approach also used in gnmi_set), it's not clear to me why you wouldn't just set xPathFlags and pbPathFlags as type []string and not use this new type.

We don't seem to use either the Set or String functionality (in gnmi_set either). Could you provide some background here (apologies if you are having to explain the codebase to me!).

gnmi_subscribe/gnmi_subscribe.go
```diff
@@ -0,0 +1,186 @@
...
+	return ""
+}
+
+func (i *arrayFlags) Set(value string) error {
+	*i = append(*i, value)
+	return nil
+}
+
+const (
+	defaultTimeout = 10
```
brianokelly1, line 44: Add a comment, what is it the default timeout for. Also include the unit in the description as for what 10 represents (seconds, minutes).

Jared-Mullin: Agreed, thank you!

PR #195, opened 2020-08-10
Added Web API stubs.

ericm:
Related issue: #194
- Added routes and view stubs for all views that will have to be implemented.

- Implemented logic and unit tests for `/prompts`. This will allow the user to store prompt answers in the backend.

### DIFF
--- a/gnxi_tester/config/web.go
+++ b/gnxi_tester/config/web.go
@@ -1,0 +1,50 @@
+/* Copyright 2020 Google Inc.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    https://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package config
+
+import (
+	"github.com/spf13/viper"
+)
+
+// Prompts represents a a prompt config set that will get stored in viper.
+type Prompts struct {
+	Name    string            `json:"name" mapstructure:"name"`
+	Prompts map[string]string `json:"prompts" mapstructure:"name"`
+	Files   map[string]string `json:"files" mapsstructure:"name"`
+}
+
+// Set prompts in viper.
+func (p *Prompts) Set() error {
+	prompts := viper.GetStringMap("web.prompts")
+	prompts[p.Name] = *p
+	viper.AllKeys()
+	if err := viper.WriteConfig(); err != nil {
+		return err
+	}
+	return nil
+}
+
+// GetPrompts returns a slice of all promts configs available.
+func GetPrompts() []Prompts {
+	out := []Prompts{}
+	webPrompts := viper.GetStringMap("web.prompts")
+	for _, p := range webPrompts {
+		if prompts, ok := p.(Prompts); ok {
+			out = append(out, prompts)
+		}
+	}
+	return out
+}
--- a/gnxi_tester/config/web.go
+++ b/gnxi_tester/config/web.go
@@ -19,7 +19,7 @@ import (
 	"github.com/spf13/viper"
 )
 
-// Prompts represents a a prompt config set that will get stored in viper.
+// Prompts represents a prompt config set that will get stored in viper.
 type Prompts struct {
 	Name    string            `json:"name" mapstructure:"name"`
 	Prompts map[string]string `json:"prompts" mapstructure:"name"`
@@ -37,7 +37,7 @@ func (p *Prompts) Set() error {
 	return nil
 }
 
-// GetPrompts returns a slice of all promts configs available.
+// GetPrompts returns a slice of all prompts configs available.
 func GetPrompts() []Prompts {
 	out := []Prompts{}
 	webPrompts := viper.GetStringMap("web.prompts")
--- a/gnxi_tester/config/web.go
+++ b/gnxi_tester/config/web.go
@@ -30,7 +30,6 @@ type Prompts struct {
 func (p *Prompts) Set() error {
 	prompts := viper.GetStringMap("web.prompts")
 	prompts[p.Name] = *p
-	viper.AllKeys()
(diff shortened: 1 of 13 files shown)


gnxi_tester/config/web.go
```diff
@@ -0,0 +1,50 @@
...
+limitations under the License.
+*/
+
+package config
+
+import (
+	"github.com/spf13/viper"
+)
+
+// Prompts represents a a prompt config set that will get stored in viper.
```
brianokelly1, line 22: nit: s/a a/a/

gnxi_tester/config/web.go
```diff
@@ -0,0 +1,50 @@
...
+	prompts := viper.GetStringMap("web.prompts")
+	prompts[p.Name] = *p
+	viper.Set("web.prompts", prompts)
+	if err := viper.WriteConfig(); err != nil {
+		return err
+	}
+	return nil
+}
+
+// GetPrompts returns a slice of all promts configs available.
```
brianokelly1, line 40: nit: prompts

gnxi_tester/config/web_test.go
```diff
@@ -0,0 +1,56 @@
...
+	viper.SetConfigFile("/tmp/config.yml")
+	t.Run("will set", func(t *testing.T) {
+		prompts := &Prompts{"test", map[string]string{}, map[string]string{}}
+		if err := prompts.Set(); err != nil {
+			t.Error(err)
+		}
+		gotMap := viper.GetStringMap("web.prompts")
+		got, ok := gotMap["test"].(Prompts)
+		if !ok {
+			t.Errorf("Couldn't find prompts in map")
```
brianokelly1, line 35: not: Whenever using errors or logs, could you ensure you use capital first letter and full stop at the end. There are a few that need fixing in this.

gnxi_tester/config/web_test.go
```diff
@@ -0,0 +1,56 @@
...
+			t.Errorf("didn't get correct prompts (-want +got) %s", diff)
+		}
+	})
+}
+func TestGetPrompts(t *testing.T) {
+	viper.SetConfigFile("/tmp/config.yml")
+	t.Run("will get", func(t *testing.T) {
+		viper.Set("web.prompts", map[string]interface{}{})
+		prompts := &Prompts{"test", map[string]string{}, map[string]string{}}
+		prompts.Set()
```
brianokelly1, line 47: It's probably still good to check Set here for an error.

gnxi_tester/config/web_test.go
```diff
@@ -0,0 +1,56 @@
...
+		got := GetPrompts()
+		if len(got) == 0 {
+			t.Error("no prompts received")
+		}
+		if diff := cmp.Diff(*prompts, got[0]); diff != "" {
```
brianokelly1, line 52: This will panic if len of got is 0.

The t.Error above logs the error but continues execution:

https://golang.org/pkg/testing/#B.Error

You probably want t.Fatal on line 50.

gnxi_tester/web/prompts.go
```diff
@@ -0,0 +1,48 @@
...
+	fmt.Fprint(w, http.StatusText(http.StatusOK))
+}
+
+// handleConfigGet will get all prompt config variables from persistent storage.
+func handlePromptsGet(w http.ResponseWriter, r *http.Request) {
+	prompts := config.GetPrompts()
+	if err := json.NewEncoder(w).Encode(prompts); err != nil {
+		logErr(r.Context(), err)
+		http.Error(w, http.StatusText(http.StatusInternalServerError), http.StatusInternalServerError)
+	}
```
brianokelly1, line 47: Do you want to also return a status code to the user here on success? Feel free to ignore if the default is to do so, then match with the above function.

gnxi_tester/web/prompts_test.go
```diff
@@ -0,0 +1,128 @@
...
+			http.StatusOK,
+			nil,
+			"[]\n",
+		},
+		{
+			"terminates correctly",
+			http.StatusOK,
+			map[string]interface{}{
+				"test": config.Prompts{
+					Name: "test",
```
brianokelly1, line 100: Add a more descriptive name here.

gnxi_tester/web/prompts_test.go
```diff
@@ -0,0 +1,128 @@
...
+			viper.SetConfigFile("/tmp/config.yml")
+			viper.Set("web.prompts", test.prompts)
+			rr := httptest.NewRecorder()
+			req, _ := http.NewRequest("GET", "/prompts", nil)
+			handler := http.HandlerFunc(handlePromptsGet)
+			handler.ServeHTTP(rr, req)
+			if code := rr.Code; code != test.code {
+				t.Errorf("wanted exit code %d but got %d", test.code, code)
+			}
+			if b, _ := ioutil.ReadAll(rr.Body); test.respBody != string(b) {
```
brianokelly1, line 123: nit: IMO, we should still check there is no error from ioutil.

gnxi_tester/web/routes.go
```diff
@@ -0,0 +1,45 @@
...
+	r.HandleFunc("/upload/{file}", handleDelete).Methods("DELETE")
+	r.HandleFunc("/run/{prompts}", handleRun).Methods("POST")
+
+	if err := http.ListenAndServe(laddr, r); err != nil {
+		log.Exit(err)
+	}
+}
+
+var logErr = func(ctx context.Context, err error) {
+	log.Errorf("error occured in view %v: %v", ctx, err)
```
brianokelly1, line 44: What does printing the context do here? Is it useful?

gnxi_tester/web/upload.go
```diff
@@ -0,0 +1,26 @@
...
+
+package web
+
+import "net/http"
+
+func handleUpload(w http.ResponseWriter, r *http.Request) {
+
+}
+
+func handleDelete(w http.ResponseWriter, r *http.Request) {
```
brianokelly1, line 24: Should this be in a separate module, if not I would change the name also. It might not be obvious to readers that upload contains both.

gnxi_tester/web/prompts.go
ericm: The framework should default the status code to 200 unless otherwise set

gnxi_tester/web/routes.go
ericm: I think ill replace it with printing the header to give context on the request

PR #196, opened 2020-08-11
gNMI Subscribe Client Functionality

Jared-Mullin:
Issue #192

### DIFF
--- a/gnmi_subscribe/gnmi_subscribe.go
+++ b/gnmi_subscribe/gnmi_subscribe.go
@@ -16,6 +16,7 @@ import (
 	"context"
 	"errors"
 	"flag"
+	"fmt"
 	"strings"
 	"time"
 
@@ -41,7 +42,7 @@ func (i *arrayFlags) Set(value string) error {
 }
 
 const (
-	defaultRequestTimeout = 10 * time.Second// This represents a value of 10 seconds and is used as a default RPC request timeout value.
+	defaultRequestTimeout = 10 * time.Second // This represents a value of 10 seconds and is used as a default RPC request timeout value.
 )
 
 var (
@@ -58,6 +59,7 @@ var (
 	encodingFormat    = flag.String("encoding", "JSON_IETF", "The encoding format used by the target for notifications")
 	suppressRedundant = flag.Bool("suppress_redundant", false, "If true, in SAMPLE mode, unchanged values are not sent by the target")
 	heartbeatInterval = flag.Uint64("heartbeat_interval", 0, "Specifies maximum allowed period of silence in seconds when surpress redundant is used")
+	updatesOnly       = flag.Bool("updates_only", false, "If true, the target only transmits updates to the subscribed paths")
 )
 
 func main() {
@@ -130,6 +132,7 @@ func main() {
 				Encoding:     pb.Encoding(encoding),
 				Mode:         subscriptionListMode,
 				Subscription: subscriptions,
+				UpdatesOnly:  *updatesOnly,
 			},
 		},
 	}
@@ -139,24 +142,93 @@ func main() {
 	}
 	switch subscriptionListMode {
 	case pb.SubscriptionList_STREAM:
-		stream()
+		if err := stream(); err != nil {
+			log.Exitf("Error using STREAM mode: %v", err)
+		}
 	case pb.SubscriptionList_POLL:
-		poll()
+		if err := poll(); err != nil {
+			log.Exitf("Error using POLL mode: %v", err)
+		}
 	case pb.SubscriptionList_ONCE:
-		once()
+		if err := once(); err != nil {
+			log.Exitf("Error using ONCE mode: %v", err)
+		}
 	}
 }
 
-func stream() {
+func stream() error {
+	for {
+		res, err := subscribeClient.Recv()
+		if err != nil {
+			return err
+		}
+		switch res.Response.(type) {
+		case *pb.SubscribeResponse_SyncResponse:
+			log.Info("SyncResponse received")
+		case *pb.SubscribeResponse_Update:
+			utils.PrintProto(res)
+		default:
+			return errors.New("Unexpected response type")
+		}
+	}
+}
+
+func poll() error {
+	ready := make(chan bool, 1)
+	ready <- true
+	pollRequest := &pb.SubscribeRequest{Request: &pb.SubscribeRequest_Poll{}}
+	if *updatesOnly {
+		res, err := subscribeClient.Recv()
+		if err != nil {
+			return err
+		}
+		if syncRes := res.GetSyncResponse(); !syncRes {
+			return errors.New("Failed to receive SyncResponse")
+		}
+		log.Info("SyncResponse received")
+	}
+	for {
+		select {
+		case <-ready:
+			log.Info("Press enter to poll")
+			fmt.Scanln()
+			subscribeClient.Send(pollRequest)
+			utils.LogProto(pollRequest)
+		default:
+			res, err := subscribeClient.Recv()
+			if err != nil {
+				return err
+			}
+			switch res.Response.(type) {
+			case *pb.SubscribeResponse_SyncResponse:
+				log.Info("SyncResponse received")
+				ready <- true
+			case *pb.SubscribeResponse_Update:
+				utils.PrintProto(res)
+			default:
+				return errors.New("Unknown response type")
+			}
+		}
+	}
 
 }
 
-func poll() {
-
-}
-
-func once() {
-	
+func once() error {
+	for {
+		res, err := subscribeClient.Recv()
+		if err != nil {
+			return err
+		}
+		switch res.Response.(type) {
+		case *pb.SubscribeResponse_SyncResponse:
+			log.Info("SyncResponse received")
+			return nil
+		case *pb.SubscribeResponse_Update:
+			utils.PrintProto(res)
+		default:
+			return errors.New("Unexpected response type")
+		}
+	}
 }
 
 func assembleSubscriptions(paths []*pb.Path) ([]*pb.Subscription, error) {
--- a/gnmi_subscribe/gnmi_subscribe.go
+++ b/gnmi_subscribe/gnmi_subscribe.go
@@ -46,7 +46,6 @@ const (
 )
 
 var (
-	subscribeClient   pb.GNMI_SubscribeClient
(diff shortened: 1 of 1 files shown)


gnmi_subscribe/gnmi_subscribe.go
```diff
@@ -139,24 +142,93 @@ func main() {
...
+	ready := make(chan bool, 1)
+	ready <- true
+	pollRequest := &pb.SubscribeRequest{Request: &pb.SubscribeRequest_Poll{}}
+	if *updatesOnly {
+		res, err := subscribeClient.Recv()
+		if err != nil {
+			return err
+		}
+		if syncRes := res.GetSyncResponse(); !syncRes {
+			return errors.New("Failed to receive SyncResponse")
```
samribeiro, line 174: Updates Only flag is set but failed to receive Sync Response first for Poll mode.

Jared-Mullin: Sounds good, changed!

samribeiro:
do you need to rebase to head to show only the latest changes?

gnmi_subscribe/gnmi_subscribe.go
```diff
@@ -77,97 +79,160 @@ func main() {
...
 	}
 	utils.PrintProto(request)
 	if err := subscribeClient.Send(request); err != nil {
 		log.Exitf("Failed to send request: %v", err)
 	}
+
 	switch subscriptionListMode {
 	case pb.SubscriptionList_STREAM:
-		stream()
+		if err := stream(subscribeClient); err != nil {
```
brianokelly1, line 125: I guess it makes sense to exit when using once. Would it make more sense to retry a few times with stream and poll given they could be very long running operations?

gnmi_subscribe/gnmi_subscribe.go
```diff
@@ -77,97 +79,160 @@ func main() {
...
+		if err != nil {
+			return err
+		}
+		switch res.Response.(type) {
+		case *pb.SubscribeResponse_SyncResponse:
+			log.Info("SyncResponse received")
+		case *pb.SubscribeResponse_Update:
+			utils.PrintProto(res)
+		default:
+			return errors.New("Unexpected response type")
```
brianokelly1, line 159: nit: no capitals at start of error string

gnmi_subscribe/gnmi_subscribe.go
```diff
@@ -77,97 +79,160 @@ func main() {
...
+		case <-ready:
+			log.Info("Press enter to poll")
+			pollInput()
+			if err := subscribeClient.Send(pollRequest); err != nil {
+				return err
+			}
+			utils.LogProto(pollRequest)
+		default:
+			res, err := subscribeClient.Recv()
+			if err == io.EOF {
```
brianokelly1, line 189: doing this will mean you won't pring the response if there is one followed by EOF.

gnmi_subscribe/gnmi_subscribe.go
```diff
@@ -77,97 +79,160 @@ func main() {
...
+	}
 
 }
 
-func once() {
-	
+func once(subscribeClient gnmi.GNMI_SubscribeClient) error {
+	for {
+		res, err := subscribeClient.Recv()
+		if err != nil {
```
brianokelly1, line 212: What if you get EOF? You checked for it in the poll function earlier.

brianokelly1: Also, all of this seems common to poll also. Could break out common code into a function to be called by both.

gnmi_subscribe/gnmi_subscribe.go
```diff
@@ -176,11 +241,55 @@ func assembleSubscriptions(paths []*pb.Path) ([]*pb.Subscription, error) {
...
 		}
 		subscriptions = append(subscriptions, subscription)
 	}
 	return subscriptions, nil
 }
+
+func subscriptionMode(subscriptionPoll, subscriptionOnce bool) (gnmi.SubscriptionList_Mode, error) {
+	switch {
+	case subscriptionPoll && subscriptionOnce:
+		return 0, errors.New("Only one of -once and -poll can be set")
```
brianokelly1, line 256: Is there a reference type you can return here rather than value? pb.SubscriptionList_UNKNOWN could be created if it doesn't exist?

gnmi_subscribe/gnmi_subscribe.go
```diff
@@ -176,11 +241,55 @@ func assembleSubscriptions(paths []*pb.Path) ([]*pb.Subscription, error) {
...
+}
+
+func parseEncoding(encodingFormat string) (gnmi.Encoding, error) {
+	encoding, ok := pb.Encoding_value[encodingFormat]
+	if !ok {
+		var encodingList []string
+		for _, name := range pb.Encoding_name {
+			encodingList = append(encodingList, name)
+		}
+		return -1, fmt.Errorf("Supported encodings: %s", strings.Join(encodingList, ", "))
```
brianokelly1, line 292: Similar to earlier,  could we create a new Encoding yupe of UNKNOWN here:

https://github.com/openconfig/gnmi/blob/964044615d9f0a475afb220fdae5026be1a75ee6/proto/gnmi/gnmi.proto#L172

gnmi_subscribe/gnmi_subscribe.go
Jared-Mullin: Sure! Changed

gnmi_subscribe/gnmi_subscribe.go
Jared-Mullin: My mistake, I should have added it, agree on common code, will refactor it

gnmi_subscribe/gnmi_subscribe.go
Jared-Mullin: I was reading [this](https://grpc.io/docs/languages/go/generated-code/) and it mentions that the end of stream is indicated by returning nil, io.EOF, since the gRPC stream guarantees order this should mean that all responses print, right?

gnmi_subscribe/gnmi_subscribe.go
```diff
@@ -176,11 +231,55 @@ func assembleSubscriptions(paths []*pb.Path) ([]*pb.Subscription, error) {
...
 		}
 		subscriptions = append(subscriptions, subscription)
 	}
 	return subscriptions, nil
 }
+
+func subscriptionMode(subscriptionPoll, subscriptionOnce bool) (gnmi.SubscriptionList_Mode, error) {
+	switch {
+	case subscriptionPoll && subscriptionOnce:
+		return UNKNOWN, errors.New("only one of -once and -poll can be set")
```
samribeiro, line 246: just return 0, it does not matter what this value is as there is a non nil error.

gnmi_subscribe/gnmi_subscribe.go
```diff
@@ -77,97 +80,149 @@ func main() {
...
-			log.Exitf("error in unmarshaling %q to gnmi Path", textPbPath)
-		}
-		pbPathList = append(pbPathList, &pbPath)
+	pbPathList, err := parsePaths(xPathFlags, pbPathFlags)
+	if err != nil {
+		log.Exitf("Error parsing paths: %v", err)
 	}
-	subscriptions, err := assembleSubscriptions(pbPathList)
+
+	subscriptions, err := assembleSubscriptions(*streamOnChange, *sampleInterval, pbPathList)
```
samribeiro, line 104: I believe you can omit the global variables from the parameters and use them directly in the function.

gnmi_subscribe/gnmi_subscribe.go
```diff
@@ -77,97 +80,149 @@ func main() {
...
-		log.Exitf("Supported encodings: %s", strings.Join(encodingList, ", "))
+	encoding, err := parseEncoding(*encodingFormat)
+	if err != nil {
+		log.Exitf("Error parsing encoding: %v", err)
 	}
 
-	var subscriptionListMode gnmi.SubscriptionList_Mode
-	switch {
-	case *subscriptionPoll && *subscriptionOnce:
+	subscriptionListMode, err := subscriptionMode(*subscriptionPoll, *subscriptionOnce)
```
samribeiro, line 93: you can omit the global variables from the parameters

samribeiro: well, unless you are testing this.

gnmi_subscribe/gnmi_subscribe.go
samribeiro: unless you are testing them =)

gnmi_subscribe/gnmi_subscribe.go
samribeiro: if stream errors out then the gRPC session is broken. The retry idea is interesting but should apply at the Dial out phase.

gnmi_subscribe/gnmi_subscribe.go
```diff
@@ -77,97 +80,149 @@ func main() {
...
+		}
+		if syncRes := res.GetSyncResponse(); !syncRes {
+			return errors.New("-updates_only flag is set but failed to receive SyncResponse first for POLL mode")
+		}
+		log.Info("SyncResponse received")
+	}
+	for {
+		select {
+		case <-ready:
+			log.Info("Press enter to poll")
```
samribeiro, line 171: this line should probably be inside the pollInput function

gnmi_subscribe/gnmi_subscribe.go
```diff
@@ -176,11 +231,55 @@ func assembleSubscriptions(paths []*pb.Path) ([]*pb.Subscription, error) {
...
+}
+
+func parseEncoding(encodingFormat string) (gnmi.Encoding, error) {
+	encoding, ok := pb.Encoding_value[encodingFormat]
+	if !ok {
+		var encodingList []string
+		for _, name := range pb.Encoding_name {
+			encodingList = append(encodingList, name)
+		}
+		return UNKNOWN, errors.New("supported encodings: " + strings.Join(encodingList, ", "))
```
samribeiro, line 282: just 0, there is non nil error, so it does not matter

gnmi_subscribe/gnmi_subscribe.go
Jared-Mullin: Agreed!

samribeiro:
testing? =)

PR #207, opened 2020-08-17
Added Subject Alternative Names to cert generation and management

### LINKED ISSUE #200: X.509 CommonName deprecation
Ref: https://golang.org/doc/go1.15#commonname
With the release of Go 1.15, treating the CommonName field on X.509 certificates as a host name when no Subject Alternative Names are present is now disabled by default. This behaviour can be re-enabled temporarily by including `GODEBUG=x509ignoreCN=0`

ericm:
Closes #200

### DIFF
--- a/certs/generate.sh
+++ b/certs/generate.sh
@@ -31,9 +31,8 @@ openssl x509 \
         -req \
         -days 365 -out ca.crt 
 
-CN="target.com"
-SUBJ="/C=NZ/ST=Test/L=Test/O=Test/OU=Test/CN=$CN"
-SAN="subjectAltName=DNS:target.com"
+SUBJ="/C=NZ/ST=Test/L=Test/O=Test/OU=Test/CN=target.com"
+SAN="subjectAltName = DNS:target.com"
 
 # Generate Target Private Key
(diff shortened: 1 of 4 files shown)


gnoi_target/gnoi_target.go
```diff
@@ -54,10 +54,10 @@ func serve() {
 	muServe.Lock()
 	defer muServe.Unlock()
 	listen, err := net.Listen("tcp", *bindAddr)
-	defer listen.Close()
```
samribeiro, line 57: thanks!

certs/generate.sh
```diff
@@ -1,15 +1,16 @@
-#!/bin/sh
+#!/bin/bash
 
 rm -f *.key *.csr *.crt *.pem *.srl
 
 SUBJ="/C=NZ/ST=Test/L=Test/O=Test/OU=Test/CN=ca"
+SAN="subjectAltName = DNS:ca"
```
samribeiro, line 6: this is not used anywhere?

ericm: That's true.

PR #229, opened 2020-08-31
Order patch

### LINKED ISSUE #226: Test Order Bug
Selecting which tests on the Web UI to run results in an error due to other tests running before provisioning operation

Jared-Mullin:
Closes #226
Adds a mat-chip-list to display the list of tests in their default order, this order can be changed and tests can be removed/added


ericm (approved):

loriscn12 (approved):

samribeiro (approved):
Thanks!

PR #234, opened 2020-09-07
gnxi_tester Coverage

### LINKED ISSUE #231: Increase test coverage for gnxi_tester
- [x]  web/target.go
- [x]  web/tests.go
- [x]  web/prompts.go
- [x]  config/config.go

Jared-Mullin:
Closes #231

### DIFF
--- a/gnxi_tester/config/history_test.go
+++ b/gnxi_tester/config/history_test.go
@@ -30,6 +30,158 @@ type Want struct {
 	err     error
 }
 
+func TestSetTarget(t *testing.T) {
+	viper.SetConfigFile("/tmp/config.yml")
+	dir, err := os.Getwd()
+	if err != nil {
+		t.Errorf("Error getting working directory: %v", err)
+	}
+	certPath := path.Join(dir, "ca.crt")
+	certKeyPath := path.Join(dir, "ca.key")
+	tests := []struct {
+		name          string
+		targetName    string
+		targetAddress string
+		targetCA      string
+		targetCAKey   string
+		config        map[string]Target
+		lastTarget    string
+		want          Want
+	}{
+		{
+			name: "No targets in history, no target specified",
+			want: Want{
+				err:     errors.New("No targets in history and no target specified"),
+				targets: map[string]Target{},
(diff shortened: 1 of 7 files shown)


gnxi_tester/config/history_test.go
```diff
@@ -30,6 +30,158 @@ type Want struct {
...
+		{
+			name: "Non-existent target",
+			config: map[string]Target{"myhost.com": {
+				Address: "localhost:9339",
+				Ca:      certPath,
+				CaKey:   certKeyPath,
+			}},
+			targetName: "nonexistenttarget",
+			lastTarget: "myhost.com",
+			want: Want{
```
loriscn12, line 84: since you're using this variable for both want and got I would probably say renaming would not be a terrible idea. First I did was trying to understand whether got:= Want{... was a mistake and I feel other may end up spending time doing the same which ... is not ideal! May I suggest "result" or something like that? Also have a comment describing what is it near definition of exported structs please.

gnxi_tester/config/history_test.go
```diff
@@ -30,6 +30,158 @@ type Want struct {
...
+			targetAddress: "localhost:9339",
+			targetCA:      certPath,
+			targetCAKey:   certKeyPath,
+			want: Want{
+				targets: map[string]Target{
+					"myhost.com": {
+						Address: "localhost:9339",
+						Ca:      certPath,
+						CaKey:   certKeyPath,
+					},
```
loriscn12, line 105: this very same content is repeated multiple times, save considerable amount of code lines by having it defined once and reused.

Jared-Mullin: Agreed! Will do!

gnxi_tester/config/history_test.go
Jared-Mullin: Agreed! Changing, thank you!

samribeiro (approved):
Thanks!

samribeiro:
needs a sync

loriscn12 (approved):

PR #236, opened 2020-09-09
Added a test to check the routes present on router

Jared-Mullin:
This test walks all routes defined and compares to a list of expected routes


samribeiro (approved):
Thanks!

loriscn12 (approved):

PR #250, opened 2020-11-25
Remove Usages of ygot/experimental/ygotutils

wenovus:
This package has been removed in more recent versions of ygot.

Resolves https://github.com/openconfig/ygot/issues/477


adibrastegarnia:
@wenovus 
Thanks. FYI: There are some other places that ygotutils is used perhaps you can address them in the same PR like this
https://github.com/google/gnxi/blob/master/gnmi/server.go#L474

samribeiro:
thank you @wenovus!

PR #257, opened 2020-12-17
Upgrade goyang and ygot dependencies

wenovus:
Also run `go mod tidy`.

There have been some recent backwards-incompatible changes, so it's good to update to the newest version.


samribeiro:
Thanks!

PR #264, opened 2021-07-14
allow loading multiple CAs into bundle

alisharif2:
The current client doesn't allow loading multiple certificates at once in a single install/rotate operation. 

New argument: `other_cas`
Comma separated list of filename prefixes that the client will use to the load the associated `.pem` and `.key` files

Usage:
```
./gnoi_cert -target_addr localhost:9339 -target_name 10.85.68.161 -logtostderr -ca device_rootCA.pem -ca_key device_rootCA.key -op provision -cert_id gnoi_demo \
-other_cas client_intermediateCA_B,client_intermediateCA_C,client_rootCA_B,client_rootCA_C,device_intermediateCA,device_rootCA -ip_address 10.85.68.161 -organization gnoi_dem -time_out 1m
```


samribeiro:
Thank you @alisharif2

PR #269, opened 2021-11-25
Oc config validate

JoseIgnacioTamayo:
Adding oc_config_validate tool

This tools allows Network Operators and Administrators test their OpenConfig-based
configurations of devices, using gNMI.


oc_config_validate/demo/certs/create_certs.sh
```diff
@@ -0,0 +1,94 @@
+#!/bin/bash
```
samribeiro, line 1: just point to: https://github.com/google/gnxi/tree/master/certs

?

JoseIgnacioTamayo: Thanks for the feedback. Ack, I now have the run_demo.sh script get and run this script from raw github.

Issue #271, opened 2021-12-01
Username and password duplication.


Joel-Cripps:
It looks like the newer gnxi/gnmi_get commands are duplicating username and passwords.

**Older client:**
go run gnmi_get.go -target_addr localhost:6032  -username admin -password 123456! -v 10  -xpath "/local-routes"  --target_name localhost -insecure -time_out 120s
```
    "metadata": {
        "_internal_repr": {
            "user-agent": [
                "grpc-go/1.30.0"
            ],
            "password": [
                "123456!"
            ],
            "username": [
                "gnmi_admin"
            ]
        },
        "flags": 0
    },
```

**New client:**
```
    "metadata": {
        "_internal_repr": {
            "user-agent": [
                "grpc-go/1.34.0"
            ],
            "username": [
                "admin",
                "admin"
            ],
            "password": [
                "123456!",
                "123456!"
            ]
        },
        "flags": 0
    },
```
    
Looks like the commit is probably : https://github.com/google/gnxi/commit/ee650d60abdcfb4ad1759b311ad0450f332e1b05
    
Changing the following line to just `return metadata.AppendToOutgoingContext(ctx)` fixes the issue.
https://github.com/google/gnxi/blob/5f44df13a119bcc3748fa15b216c38bb9072a1d8/utils/credentials/credentials.go#L168

I'm not sure if there is any reason to be sending multiple username/passwords, or have username and passwords as lists instead of a string.

wenovus:
I wasn't able to reproduce when running your command and just printing out the metadata:
```go
129   ctx = credentials.AttachToContext(ctx)                                                                                                          
130                                                                                                                                                   
131   md, _ := metadata.FromOutgoingContext(ctx)                                                                                                      
132   fmt.Println(md)                                                                                                                                 
```
I'm failing to see what might be causing `AttachToContext` to be called twice on the same `ctx` object -- is your code being run in parallel?

Can you try if https://github.com/google/gnxi/pull/273 fixes it?

samribeiro:
assuming fixed in #273

PR #273, opened 2021-12-18
Override username and password instead of appending them.

wenovus:
This ensures that the same context being called twice by `AttachToContext` doesn't replicate the username and password metadata.

Possibly addresses #271.


samribeiro:
thanks!

PR #289, opened 2022-01-12
Add model_data field to subscribe client

GeorgeH049:
To support submission of the use_models field via subscribe RPC as per spec


samribeiro:
thanks!

PR #295, opened 2022-03-10
Support for -prefix and -data_type in gnmi_get and -prefix for gnmi_set.

brownoldmonk:
1. Add -prefix parameter to gnmi_set.
2. Add -prefix and -data_type to gnmi_get.

 data_type can be 0 (ALL), 1 (CONFIG), 2 (STATE) 3 (OPERATIONAL). Default is 0.

 prefix is a string that can be used in GNMI Get/Set messages. Default is none.


samribeiro:
Thanks!

PR #325, opened 2023-03-11
Code improvements for oc_config_validate

JoseIgnacioTamayo:
* Move the validation of the Target values to where they are defined (in Context)
 * Use a ContextManager to ensure gNMI Connection and Close to the Target
 * Factorize code that sends gNMI Gets and asserts


xavier-contreras:
LGTM

PR #334, opened 2023-05-07
oc_config_validate streaming on_change tests

JoseIgnacioTamayo:
Add support for Stream OnChange Subscription tests.

2 TestClasses, which verify that an OnChange subscription gets Updates with different values.


xavier-contreras:
LGTM

PR #335, opened 2023-05-07
oc_config_validate streaming fixes

JoseIgnacioTamayo:
Some code refactoring and better Exception catching.

The functional change is that STREAM subscriptions are valid if a sync_response messages is received, this is no longer an optional verification (https://github.com/openconfig/reference/blob/master/rpc/gnmi/gnmi-specification.md#3523-sending-telemetry-updates)


xavier-contreras:
LGTM

PR #380, opened 2024-06-06
Fix bug in type comparison.

xavier-contreras:
Fixes bug introduced in pull/377


JoseIgnacioTamayo:
Oh.. this was my missed. I corrected this line to fix a pylint error, and of course I forgot the '!='...

Sorry. Too bad this was not caught by the demo tests.

PR #381, opened 2024-06-14
[oc_config_validate] Generate custom wifi_arista binding

xavier-contreras:
@JoseIgnacioTamayo FYI, please review.

This updates the shell script in the following manner:

- Adds some verbosity to improve UX
- Downloads arista-wifi-augments.yang and creates wifi_arista.py so the openconfig model includes the vendor augments

This PR includes a fresh run of the script and thus the updated init and new binding.


JoseIgnacioTamayo:
LGTM.

I would suggest we make a small increase in the version, 2.6.3, so signal this contains a custom binding Python file.

Also, it would be great to have some demonstration of the use of the custom model, likely with a testfile like oc_config_validate/tests/ap.yaml. There it would be visible how to call the custom binding and a few interesting Leafs/Containers.

xavier-contreras:
> LGTM.
> 
> I would suggest we make a small increase in the version, 2.6.3, so signal this contains a custom binding Python file.
> 
> Also, it would be great to have some demonstration of the use of the custom model, likely with a testfile like oc_config_validate/tests/ap.yaml. There it would be visible how to call the custom binding and a few interesting Leafs/Containers.

Thanks!  I'll include the version update before merging, and will do the documentation parts you suggest in a separate PR.

JoseIgnacioTamayo:
Ack!
LGTM again.

PR #403, opened 2026-03-30
Update OC models to v5.6.0

chechoRP:
* Update OC Models to version 5.6.0
* Bump oc_config_validate version to 2.11.0


chechoRP:
Hi @JoseIgnacioTamayo @xavier-contreras PTAL at this PR to bump OC version. Thanks!

oc_config_validate/oc_config_validate/__init__.py
xavier-contreras, line 1: This version is for the framework itself, we only change it when oc_config_validate itself changes.  For model updates, we don't.

xavier-contreras (changes requested):
Comment on the version.

oc_config_validate/oc_config_validate/__init__.py
chechoRP: Fixed.

PR #404, opened 2026-05-14
gnoi/os: limit OS image size in ReceiveOS to prevent unbounded memory allocation 

kazisabu:
Added a maximum OS image size limit to ReceiveOS to prevent unbounded
memory allocation (DoS).

Changes made in gnoi/os/server.go:

1. Added const maxOSImageSize = 1 << 30 (1GB) constant after
   receiveChunkSizeAck variable declaration (lines 29-31).

2. Added a size check after bb.Write() in the TransferContent case
   of ReceiveOS that returns an error when the buffer exceeds
   maxOSImageSize (lines 186-188).

Without this fix, an authenticated client can stream unlimited chunks
to the Install RPC, causing the server to allocate memory until
OOM-killed. The existing installToken semaphore limits concurrency
to 1 but does not prevent a single session from exhausting memory.

### DIFF
--- a/gnoi/os/server.go
+++ b/gnoi/os/server.go
@@ -25,6 +25,9 @@ import (
 )
 
 var receiveChunkSizeAck uint64 = 12000000
+// maxOSImageSize limits the maximum size of an OS image accepted by ReceiveOS
+// to prevent unbounded memory allocation (DoS).
+const maxOSImageSize = 1 << 30 // 1GB
 
 // Server is an OS Management service.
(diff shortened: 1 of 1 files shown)


kazisabu:
For the record, this issue has been validated as a security vulnerability by the Google VRP team. I am currently in the process of requesting a CVE ID for this issue to ensure it is documented for the community. If any maintainers are available to review the fix, I would appreciate the feedback.

